PR #8, opened 2017-07-27
Make book always present as Device16x9Portrait (BL-4821)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/8)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
@@ -17,6 +17,7 @@ import java.io.FileFilter;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 import org.jsoup.Jsoup;
 import org.jsoup.nodes.Document;
@@ -26,6 +27,8 @@ import org.sil.bloom.reader.models.Book;
 
 
 public class ReaderActivity extends AppCompatActivity {
+
+    private static final Pattern sLayoutPattern = Pattern.compile("\\S+([P|p]ortrait|[L|l]andscape)$");
 
     private ViewPager mPager;
     private BookPagerAdapter mAdapter;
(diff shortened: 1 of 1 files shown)


hatton:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 126 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/8#-Kq3vFq2Y4zUDb0-urpg:-Kq3vFq2Y4zUDb0-urph:b-hxnojt) ([raw file](https://github.com/bloombooks/bloomreader/blob/6f46351dfd7a557df8e2f47e48b7fd0e29679a7b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L126)):*
> ```Java
>     private void modifyLayout(Element page, String newLayout) {
>         for (String className : page.classNames()) {
>             if (sLayoutPattern.matcher(className).matches()) {
> ```

What are the situations where it does end in one of these? And in those situations, why don't we want to force it to the device layout?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/8)*

andrew-polk:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 126 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/8#-Kq3vFq2Y4zUDb0-urpg:-Kq3w-FbPHpYn4mC3vn2:bspcenk) ([raw file](https://github.com/bloombooks/bloomreader/blob/6f46351dfd7a557df8e2f47e48b7fd0e29679a7b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L126)):*

What are the situations where it does end in one of these? And in those situations, why don't we want to force it to the device layout?
</blockquote></details>

This is looking at each class name; so it includes page-left and bloom-page, for example.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/8)*

andrew-polk:
*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 126 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/8#-Kq3vFq2Y4zUDb0-urpg:-Kq3wGH8D1C8IX0qZyzr:bbpugaw) ([raw file](https://github.com/bloombooks/bloomreader/blob/6f46351dfd7a557df8e2f47e48b7fd0e29679a7b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L126)):*

This is looking at each class name; so it includes page-left and bloom-page, for example.
</blockquote></details>

side-left, I mean

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/8)*

PR #10, opened 2017-08-01
BL-4832 Make Get Books from Wifi a dialog-launching command

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/10)


JohnThomson:
Implemented in https://github.com/BloomBooks/BloomReader/pull/10.

Note that the first of two commits in this makes the command launch a new first-class activity which fills the screen, rather than one that looks like a dialog. This might be a useful point to return to if we decide we want that. However, some comments and other improvements added later might need to survive.

andrew-polk:
A couple niggles, otherwise :lgtm: 

---

Reviewed 14 of 14 files at r1.

---

*[app/src/main/res/values/strings.xml, line 28 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/10#-KqUMVDnCh0TuWRRZllq:-KqUMVDnCh0TuWRRZllr:b-knmtix) ([raw file](https://github.com/bloombooks/bloomreader/blob/e3250a85e426f89b6e79f30a6b9ea1f2563d8fba/app/src/main/res/values/strings.xml#L28)):*
> ```XML
>     <string name="downloading">Downloading...</string>
>     <string name="looking_for_adds">Looking for book advertisements on wifi network %1s...</string>
>     <string name="no_wifi_connected">The phone does not appear to be connected to a wifi network.</string>
> ```

I think we are opting for "device" rather than "phone".
And BloomDesktop uses "WiFi", I believe.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/10#-:-KqUMhV_z8nC9K8XBKXD:b-47t2i7)*

JohnThomson:
Fixed...also pushed a few comment improvements I had previously made but not committed.

---

Review status: 11 of 14 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/res/values/strings.xml, line 28 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/10#-KqUMVDnCh0TuWRRZllq:-KqY7p17U6I5uS4xlTjV:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/e3250a85e426f89b6e79f30a6b9ea1f2563d8fba/app/src/main/res/values/strings.xml#L28)):*

I think we are opting for "device" rather than "phone".
And BloomDesktop uses "WiFi", I believe.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/10#-:-KqY7r5Otn5Hwaw4PEh7:bw5ygw7)*

andrew-polk:
Reviewed 3 of 3 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/10)*

PR #12, opened 2017-08-02
Sort book list alphabetically (BL-4833)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/12)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/models/Book.java
+++ b/app/src/main/java/org/sil/bloom/reader/models/Book.java
@@ -1,6 +1,7 @@
 package org.sil.bloom.reader.models;
 
 import java.io.File;
+import java.util.Comparator;
 
 public class Book {
     public static final String BOOK_FILE_EXTENSION = ".bloomd";
@@ -18,5 +19,17 @@ public class Book {
     @Override
     public String toString() {
         return this.name;
+    }
+
+    public static final Comparator<Book> AlphabeticalComparator = new AlphabeticalBookComparator();
+
+    public static class AlphabeticalBookComparator implements Comparator<Book> {
+        @Override
+        public int compare(Book book1, Book book2) {
+            int nameCompare = book1.name.compareToIgnoreCase(book2.name);
+            if (nameCompare != 0)
+                return nameCompare;
+            return book1.path.compareToIgnoreCase(book2.path);
+        }
     }
 }
--- a/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java
+++ b/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java
@@ -12,6 +12,7 @@ import java.io.File;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 public class BookCollection {
@@ -39,6 +40,7 @@ public class BookCollection {
     private Book addBook(String path) {
         Book book = new Book(path);
         _books.add(book);
+        Collections.sort(_books, Book.AlphabeticalComparator);
         return book;
     }
 


JohnThomson:
Some of what we're doing here is starting to creep away from the UI. Is it time to think about unit testing?

---

Reviewed 2 of 2 files at r1.

---

*[app/src/main/java/org/sil/bloom/reader/models/Book.java, line 32 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuEjxXcRdGVpCm73U:-KqZuEjxXcRdGVpCm73V:b-ima2qi) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/Book.java#L32)):*
> ```Java
>             if (nameCompare != 0)
>                 return nameCompare;
>             return book1.path.compareToIgnoreCase(book2.path);
> ```

Aren't they all kept in one folder? In which case this extra step won't change the result you already had.

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 43 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuLEil0fczTXrzhj2:-KqZuLEil0fczTXrzhj3:b-v5puws) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L43)):*
> ```Java
>         Book book = new Book(path);
>         _books.add(book);
>         Collections.sort(_books, Book.AlphabeticalComparator);
> ```

Don't we also need to sort the whole collection when we first load it? (Or, if the initial load calls addBook for each one, sorting every time is pretty inefficient.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-:-KqZufBQkZuDtCIne5Nl:b-byu19d)*

andrew-polk:
Yes, it may be time to start thinking about unit testing. I'll look into it. If 4.0 was ready, I would say it wasn't worth the overhead of setting it all up yet for MVP1. But BR is basically useless without a 4.0 beta.

---

---

*[app/src/main/java/org/sil/bloom/reader/models/Book.java, line 32 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuEjxXcRdGVpCm73U:-KqbugIk6V80eR6kVWpv:b33yxaf) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/Book.java#L32)):*

Aren't they all kept in one folder? In which case this extra step won't change the result you already had.
</blockquote></details>

Currently, yes. And currently we assume name is unique. This is future-proofing.

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 43 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuLEil0fczTXrzhj2:-Kqbumyr2A9Dx2Xk6iGG:b-vcvc7k) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L43)):*

Don't we also need to sort the whole collection when we first load it? (Or, if the initial load calls addBook for each one, sorting every time is pretty inefficient.)
</blockquote></details>

Yes, we inefficiently sort for each on initial load. Every time I bring up efficiency in reviews, I'm always shot down with the, "why introduce more complexity when we don't know if it matters" argument, so this time I decided to code that way. If you see this case as different, I would like to understand why.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-:-KqbuNkBO_IhMFpxM2iC:b-7h95xz)*

JohnThomson:
Doesn't have to be part of this PR...just a thought.

---

---

*[app/src/main/java/org/sil/bloom/reader/models/Book.java, line 32 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuEjxXcRdGVpCm73U:-KqcMxFrvYCKx5ZOVGP6:b2nhktp) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/Book.java#L32)):*

Currently, yes. And currently we assume name is unique. This is future-proofing.
</blockquote></details>

OK.

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 43 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuLEil0fczTXrzhj2:-KqcN-QU744W7eMSCp2y:b-olcjli) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L43)):*

Yes, we inefficiently sort for each on initial load. Every time I bring up efficiency in reviews, I'm always shot down with the, "why introduce more complexity when we don't know if it matters" argument, so this time I decided to code that way. If you see this case as different, I would like to understand why.
</blockquote></details>

Good question. Premature optimization is certainly a danger. I tend to think it's worth a little thought especially if something more efficient does NOT add to complexity. For example, in C# I would make it a sorted collection..I believe Java has java.util.PriorityQueue. That lets you add n items in n log n time and might actually be simpler, whereas the current code is going to be n-squared log n, which grows pretty fast...400k or so string comparisons for 200 items.

It can stand for MVP. Won't be hard to fix if we need to. The main thing was to be sure it DOES get sorted initially.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-:-KqcMt7FwaQAr6cILLJY:b-72936z)*

andrew-polk:
*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 43 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/12#-KqZuLEil0fczTXrzhj2:-Kqd68qIy-LYAzE6AqMU:br6lm8k) ([raw file](https://github.com/bloombooks/bloomreader/blob/3de53e04249a1f4f08c4a0f4c946d8aa38499f88/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L43)):*

Good question. Premature optimization is certainly a danger. I tend to think it's worth a little thought especially if something more efficient does NOT add to complexity. For example, in C# I would make it a sorted collection..I believe Java has java.util.PriorityQueue. That lets you add n items in n log n time and might actually be simpler, whereas the current code is going to be n-squared log n, which grows pretty fast...400k or so string comparisons for 200 items.

It can stand for MVP. Won't be hard to fix if we need to. The main thing was to be sure it DOES get sorted initially.
</blockquote></details>

Yes, a sorted collection should have worked. I tried that initially, but was having trouble with it. I don't remember why. I didn't actually find the PriorityQueue, though, so maybe I was just using the wrong data structure. I'll look at it. Thanks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/12)*

PR #36, opened 2017-09-06
BL-4955 push analytics information to Segment.io

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/36)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -103,6 +103,7 @@ dependencies {
     compile 'com.android.support:design:23.4.0'
     // Supports the WiFi module's 'server' used to receive books from Bloom desktop
     compile 'cz.msebera.android:httpclient:4.4.1.2'
+    compile 'com.segment.analytics.android:analytics:4.+'
     compile 'commons-io:commons-io:2.4'
     testCompile 'junit:junit:4.12'
 }
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -4,6 +4,8 @@
 
     <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <!-- Used at least for analytics, soon for Bloom Library download. Was present when I added
+    analytics, so may be used for something else also. -->
     <uses-permission android:name="android.permission.INTERNET" />
     <!-- currently just because it's how we get the device name for WiFi, but eventually we will have
     other Bluetooth functions. -->
--- a/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java
+++ b/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java
@@ -2,6 +2,15 @@ package org.sil.bloom.reader;
 
 import android.app.Application;
 import android.content.Context;
+import android.content.pm.PackageManager;
+import android.os.Build;
+import android.widget.Toast;
+
+import com.segment.analytics.Analytics;
+
+import org.sil.bloom.reader.models.BookCollection;
+
+import java.io.File;
 
 // Our special Application class which allows us to share information easily between activities.
 // Note that anything stored here should not be expected to persist indefinitely.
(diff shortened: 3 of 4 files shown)


andrew-polk:
Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[app/build.gradle, line 106 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-KtOynGud_vDzgK4htSl:-KtOynGud_vDzgK4htSm:b-787kcz) ([raw file](https://github.com/bloombooks/bloomreader/blob/a6ec0763c29026cd5ca119bf3392ae58004bedc5/app/build.gradle#L106)):*
> ```Gradle
>     // Supports the WiFi module's 'server' used to receive books from Bloom desktop
>     compile 'cz.msebera.android:httpclient:4.4.1.2'
>     compile 'org.apache.directory.studio:org.apache.commons.io:2.4'
> ```

This is the one I was trying to use which caused our alpha to not be available for "any devices". I replaced it with the commons-io one below which made everything happy. I assume you can just remove this line now.

---

*[app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java, line 39 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-KtOzijhtC475RlfAitl:-KtOzijhtC475RlfAitm:b-3wgmey) ([raw file](https://github.com/bloombooks/bloomreader/blob/a6ec0763c29026cd5ca119bf3392ae58004bedc5/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java#L39)):*
> ```Java
>         } else {
>             if (BuildConfig.FLAVOR == "production") {
>                 writeKey = "EfisyNbRjBYIHyHZ9njJcs5dWF4zabyH"; // Source BloomReader
> ```

What prevents someone from picking this up in our source code and using it to pollute our analytics?
I realize the same question probably applies to our other apps...

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 89 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-KtP-GyDKun8KvATqsPa:-KtP-GyDKun8KvATqsPb:b-wr72pd) ([raw file](https://github.com/bloombooks/bloomreader/blob/a6ec0763c29026cd5ca119bf3392ae58004bedc5/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L89)):*
> ```Java
>     @Override
>     protected void onPause() {
>         if (isFinishing() && mAdapter.mThisPageIndex > 1) {
> ```

If the idea is to see how much of a book is read, it seems like we would want to track when they close a book without reading it, too.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/36)*

JohnThomson:
Review status: 2 of 4 files reviewed at latest revision, 3 unresolved discussions.

---

*[app/build.gradle, line 106 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-KtOynGud_vDzgK4htSl:-KtRZUmc6WK5cuAcOXFe:bmee131) ([raw file](https://github.com/bloombooks/bloomreader/blob/a6ec0763c29026cd5ca119bf3392ae58004bedc5/app/build.gradle#L106)):*

This is the one I was trying to use which caused our alpha to not be available for "any devices". I replaced it with the commons-io one below which made everything happy. I assume you can just remove this line now.
</blockquote></details>

Done. (Glad you knew that...I wondered if they might be more-or-less the same thing.)

---

*[app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java, line 39 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-KtOzijhtC475RlfAitl:-KtRZbE7QKv0mtRyfVpb:b-pb73xv) ([raw file](https://github.com/bloombooks/bloomreader/blob/a6ec0763c29026cd5ca119bf3392ae58004bedc5/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java#L39)):*

What prevents someone from picking this up in our source code and using it to pollute our analytics?
I realize the same question probably applies to our other apps...
</blockquote></details>

I don't know of any way to prevent it...Bloom desktop has its key right in the source...worry if it becomes a problem, maybe? It's not obvious how we could improve things.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 89 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-KtP-GyDKun8KvATqsPa:-KtR_FcqFDRS9wlE1b-A:bvt5n7l) ([raw file](https://github.com/bloombooks/bloomreader/blob/a6ec0763c29026cd5ca119bf3392ae58004bedc5/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L89)):*

If the idea is to see how much of a book is read, it seems like we would want to track when they close a book without reading it, too.
</blockquote></details>

I was thinking in terms of counting pages read (but without generating an event for every page flip). So if they didn't read any, there's nothing to report. Of course arguably they probably haven't really done any reading unless they get past the xmatter. It's a crude approximation at best. I can take it out.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/36)*

andrew-polk:
:lgtm:
But this is the kind of thing @Hatton probably wants to take a peak at from a high-level what-do-we-want-to-report perspective.

---

Reviewed 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/36#-:-KtRbOo3XS0oq6gGHcVC:b1rhj98)*

PR #38, opened 2017-09-18
Bl 4991 share bloom books

rjconrad:
Adds the option to share the bloomd file with other apps to the Book context menu.

Improves the book import that happens when a bloomd file is opened in another app and referred to BloomReader.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/38)


rjconrad:
The change to the README is adding an apostrophe to the word "world" in the phrase "create books in the world's thousands of minority languages"

andrew-polk:
Looks good overall. A few questions (mostly my ignorance) and a few nitpicks.

---

Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 88 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRuy4eQYZMXPgGTlVR:-KuRuy4fdOoe21mOWeMb:b-ye11m) ([raw file](https://github.com/bloombooks/bloomreader/blob/823b739b31a005b020dafa7535261ec7f304c337/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L88)):*
> data.getPath().toLowerCase().endsWith(Book.BOOK_FILE_EXTENSION)

Looks like we lost this check. Or do we now need to try regardless and let ensureBookIsInCollection handle both cases?

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 190 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRvJh0cOmSGp5tOo0O:-KuRvJh0cOmSGp5tOo0P:bctwrjk) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L190)):*
> ```Java
> 
>     private void shareBook(){
>         int position = mListView.getCheckedItemPosition();
> ```

Not a huge deal, but we could extract these two lines and the first two in DeleteBook to a common method.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 193 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRwTrZH0oqQ-Txeazp:-KuRwTrZH0oqQ-Txeazq:b-2hk0bh) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L193)):*
> "org.sil.bloom.reader.fileprovider"

Is there any way to get this string from the class or some other way to keep from hardcoding it?

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 199 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRviOsf1DBi-o6laoR:-KuRviOsf1DBi-o6laoS:b-blddhc) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L199)):*
> ```Java
>         shareIntent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
>         shareIntent.putExtra(Intent.EXTRA_STREAM, fileUri);
>         shareIntent.setType("*/*");
> ```

Is there any way to register a new type when Bloom Reader installs so that we can actually send that new type and the receiver will know it is a file for Bloom Reader?

Or do we already so something like that? 
Ideal user experience: receiver gets the book loaded into Bloom Reader as soon as it is received (not sure if this is possible). What do they have to do currently? Simply open the file?

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 130 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuS-UmdMI84dUkG-LfG:-KuS-UmdMI84dUkG-LfH:bflb9be) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L130)):*
> ```Java
>         String name = nameOfBloomFile(context, bookUri);
>         if(!name.endsWith(Book.BOOK_FILE_EXTENSION))
>             return null; //I'm not 100% sure this won't weed out some legit book files
> ```

Help me understand more. How could it?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-:-KuS0EeYdMHZcHCPpPIh:b8g9tqp)*

rjconrad:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 88 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRuy4eQYZMXPgGTlVR:-KuVHX2pqiVAnQYAnC5g:b-a8b0hd) ([raw file](https://github.com/bloombooks/bloomreader/blob/823b739b31a005b020dafa7535261ec7f304c337/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L88)):*

> data.getPath().toLowerCase().endsWith(Book.BOOK_FILE_EXTENSION)

Looks like we lost this check. Or do we now need to try regardless and let ensureBookIsInCollection handle both cases?
</blockquote></details>

This check got moved into ensureBookIsInCollecction. If the nameOfBloomFile() doesn't end with the .bloomd extension, ensureBookIsInCollection returns null.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 190 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRvJh0cOmSGp5tOo0O:-KuVI6olXtWJNPoWVk57:b8bof47) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L190)):*

Not a huge deal, but we could extract these two lines and the first two in DeleteBook to a common method.
</blockquote></details>

Sure, I can do that.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 193 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRwTrZH0oqQ-Txeazp:-KuVI9JPyKOd7XsJMy4s:bcsqw4j) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L193)):*

> "org.sil.bloom.reader.fileprovider"

Is there any way to get this string from the class or some other way to keep from hardcoding it?
</blockquote></details>

It has to match the name that's in the provider tag of the Manifest. The documentation did not mention any way to get it from that source in the Java.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 199 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRviOsf1DBi-o6laoR:-KuVIw43SLs4soeslZ_V:b4raaci) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L199)):*

Is there any way to register a new type when Bloom Reader installs so that we can actually send that new type and the receiver will know it is a file for Bloom Reader?

Or do we already so something like that? 
Ideal user experience: receiver gets the book loaded into Bloom Reader as soon as it is received (not sure if this is possible). What do they have to do currently? Simply open the file?
</blockquote></details>

This isn't a way to share directly between two devices, but a way to share the file with another app on the phone that could be used to get it to somebody else. That could be emailing it, using the phone's bluetooth utility or a file explorer app to save it on the sd card (although that's where the files are stored by default anyway). 

When the receiver taps on the bloomd file (in the email for instance), they're prompted to open BloomReader because we have that extension registered in the intent filters.

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 130 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuS-UmdMI84dUkG-LfG:-KuVKVH1SEC_v4hc5m_N:b-sf7fle) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L130)):*

Help me understand more. How could it?
</blockquote></details>

What I ran into was that when I selected the bloomd file in three different apps, the intent that BloomReader received was formatted three slightly different ways. The file explorer app did it the simple way that we already had code for. Gmail did it the way that's documented. And evernote did it similarly to gmail but the filename was stored in a different field in the cursor.

So if a different app puts the filename somewhere else that I don't know about, we won't know that the extension is bloomd. It might be better to try to open it, and see if it works, but I wasn't sure how to write that code. Also we wouldn't necessarily know the right name for the book (unless it's somewhere inside the file also).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/38)*

JohnThomson:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 130 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuS-UmdMI84dUkG-LfG:-KuVT1ehAEEyvawZQhIk:b-b1rhcn) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L130)):*

What I ran into was that when I selected the bloomd file in three different apps, the intent that BloomReader received was formatted three slightly different ways. The file explorer app did it the simple way that we already had code for. Gmail did it the way that's documented. And evernote did it similarly to gmail but the filename was stored in a different field in the cursor.

So if a different app puts the filename somewhere else that I don't know about, we won't know that the extension is bloomd. It might be better to try to open it, and see if it works, but I wasn't sure how to write that code. Also we wouldn't necessarily know the right name for the book (unless it's somewhere inside the file also).
</blockquote></details>

More fundamentally, if you can't actually find the file that the intent is asking you to open, you can't copy it to the bloom folder nor try to open it.
If you had some way to open the file without knowing its name, it would be a pretty good check that it's really a bloom book to attempt to unzip it, look for the expected version.txt file, and maybe a couple of other characteristic bloom files like basePage.css, and look for a single HTM file. The name of that file would be a reasonable book name.
But, I don't see how to open it without tracking down the name!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/38)*

rjconrad:
The path and name that come from a ContentProvider are different from the
file's actual location on disk and its original name. (This has to do with
permissions so that only the intended app has access to the file being
 shared). There is a way to get the original name according to the API, but
in at least one case this was followed inconsistently.

rjconrad:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/models/BookCollection.java, line 130 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuS-UmdMI84dUkG-LfG:-KuVgevjAU4I2_Uius3u:bmbqgz9) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/models/BookCollection.java#L130)):*
(Sorry for the repeat. I tried replying by email and it went to the wrong thread.)
> The path and name that come from a ContentProvider are different from the file's actual location on disk and its original name. (This has to do with permissions so that only the intended app has access to the file being  shared). There is a way to get the original name according to the API, but in at least one case this was followed inconsistently.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/38)*

rjconrad:
Review status: 5 of 11 files reviewed at latest revision, 4 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 190 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-KuRvJh0cOmSGp5tOo0O:-KuysrD2nhBZzhgAX1_j:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/28df153bea795884d2ab188d546fd5e1656fcf2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L190)):*

Sure, I can do that.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/38)*

rjconrad:
I now have it validating incoming files by making sure we can unzip them and that there is an html file inside.

andrew-polk:
I'm not going to pretend I understood it all, but nothing looks wrong. I'm waiting to see if JH wants a look before we merge.

---

Reviewed 6 of 6 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/38#-:-Kv3kErw-rCSS-otBrY3:b-iz95nw)*

PR #42, opened 2017-10-04
User must initiate all page turns (BL-5067)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/42)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
@@ -122,13 +122,18 @@ public class ReaderActivity extends BaseActivity {
     final int MIN_PAGE_SWITCH_MILLIS = 3000;
 
     // This is the routine that is invoked as a result of a call-back from javascript indicating
-    // that the current page is complete (which in turn typically follows a notifcation from
+    // that the current page is complete (which in turn typically follows a notification from
     // Java that a sound finished playing...but it is the JS that knows it is the last audio
     // on the page). Basically it is responsible for flipping to the next page (see goToNextPageNow).
     // But, we get the notification instantly if there is NO audio on the page.
     // So we do complicated things with a timer to make sure we don't flip the page too soon...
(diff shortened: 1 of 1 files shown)


hatton:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 134 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/42#-KvcPQMfIsHoJw8Up1G3:-KvcPQMfIsHoJw8Up1G4:b-6dq41u) ([raw file](https://github.com/bloombooks/bloomreader/blob/66d04823c5ef1db57add0070b7581e79b5a6d43b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L134)):*
> ```Java
> 
>         // For now, we have decided we want to force the user to initiate page turning always (BL-5067).
>         if (true)
> ```

Not clear what the if(true) does, but... if that's what you want ;-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/42)*

andrew-polk:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 134 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/42#-KvcPQMfIsHoJw8Up1G3:-KvcQi1BmBGHdPiKa_6e:b-6wt30s) ([raw file](https://github.com/bloombooks/bloomreader/blob/66d04823c5ef1db57add0070b7581e79b5a6d43b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L134)):*

Not clear what the if(true) does, but... if that's what you want ;-)
</blockquote></details>

Just lets me keep the code in a IDE-compiled state so it doesn't get hopelessly lost/out of date. Java doesn't just let you return like C#. It is a compile error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/42)*

PR #59, opened 2017-10-26
Success/Failure toasts for initial load of Device ID JSON.

rjconrad:
During application onCreate, if there is no stored Device Id in the SharedPrefs, we check for a json file. If we succeed in parsing the file, there's a success toast. If we find a file but fail to parse it, there's a failure toast. Once device id metadata is loaded, we never look at the json again. This means changing the metadata requires a reinstall.

For the people loading the json file, once the file is transferred, they need to make sure BR is closed. Then they should run it and watch for the success message.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/59)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java
+++ b/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java
@@ -2,8 +2,8 @@ package org.sil.bloom.reader;
 
 import android.app.Application;
 import android.content.Context;
+import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
-import android.support.annotation.Nullable;
 import android.util.Log;
 import android.widget.Toast;
 
@@ -23,6 +23,8 @@ import java.io.File;
 public class BloomReaderApplication extends Application {
     public static final String SHARED_PREFERENCES_TAG = "org.sil.bloom.reader.prefs";
     public static final String LAST_RUN_BUILD_CODE = "lastRunBuildCode";
+    public static final String ANALYTICS_DEVICE_PROJECT = "analyticsDeviceGroup";
+    public static final String ANALYTICS_DEVICE_ID = "analyticsDeviceId";
     public static final String DEVICE_ID_FILE = "deviceId.json";
 
     private String bookToHighlight;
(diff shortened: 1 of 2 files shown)


JohnThomson:
It would be pretty easy to make it check the json each time. But changing the device ID is something of a big deal...analytics will presumably report it as a different device, and confusingly, we would never get analytics saying Bloom was installed on that device. So you're probably right to make it quite hard to change. @hatton, what do you think?

---

Reviewed 2 of 2 files at r1.

---

*[app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java, line 112 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/59#-KxNuWcpdy_XdonDQABC:-KxNuBKdlCS3LNM69Q4R:bosz4nd) ([raw file](https://github.com/bloombooks/bloomreader/blob/6e6ff7d6677d4e521b6901d4b50cc550bd2fabfb/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java#L112)):*
> ```Java
>         }
>         catch (JSONException e){
>             Log.e("Analytics", "Error processing deviceId file json.");
> ```

Let's include the exception message in the log. May give useful clues to what is wrong.

---

*[app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java, line 123 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/59#-KxNtZaG44r3PIBsjSBv:-KxNtZaG44r3PIBsjSBw:b-3lamf8) ([raw file](https://github.com/bloombooks/bloomreader/blob/6e6ff7d6677d4e521b6901d4b50cc550bd2fabfb/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java#L123)):*
> ```Java
> 
>     private static void reportDeviceIdParseSuccess(String project, String device){
>         String successMessage = getBloomApplicationContext().getString(R.string.metadata_loaded) + "\nProject: " + project + "\nDevice: " + device;
> ```

This feels half-hearted. If we're going to make this localizable, the whole message should be, and we should use an appropriate Format routine to inject the project and device where the localizer wants them (with localized labels). Since we only expect this toast to be seen by (very likely English-speaking) technicians, not localizing any of it would also make sense.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/59#-:-KxNvNsAKrVpD3jLXreb:bygsb2p)*

rjconrad:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java, line 123 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/59#-KxNtZaG44r3PIBsjSBv:-KxrFJEWR4w6rJGZlMP0:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/6e6ff7d6677d4e521b6901d4b50cc550bd2fabfb/app/src/main/java/org/sil/bloom/reader/BloomReaderApplication.java#L123)):*

This feels half-hearted. If we're going to make this localizable, the whole message should be, and we should use an appropriate Format routine to inject the project and device where the localizer wants them (with localized labels). Since we only expect this toast to be seen by (very likely English-speaking) technicians, not localizing any of it would also make sense.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/59)*

JohnThomson:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/59#-:-KxwyYjGTjqvWE9442WH:bnfp4nl)*

JohnThomson:
Reviewed 1 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/59)*

PR #69, opened 2017-11-06
[BL-5088] Update Play Store screenshots.

rjconrad:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/69)


andrew-polk:
@rjconrad, please go ahead and base this on the Version1.0 branch.

JohnThomson:
These look basically good to me, but I'm not sure exactly what we wanted to correct in the old ones. The two shots from "moon and cap" look very similar...if we're going to have three, I wonder whether there's any other aspect of the program we could usefully capture? I don't think there's anything very interesting in the menu pages...what about a page with a more complex layout? Things like side-by-side pictures and text are one thing that doesn't work in epubs but does in BR.

PR #72, opened 2017-11-09
[BL-5216] Use slide transition between activities

rjconrad:
Here's what I have, but I don't think the effect looks very nice...

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/72)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/MainActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/MainActivity.java
@@ -1,12 +1,14 @@
 package org.sil.bloom.reader;
 
 import android.Manifest;
+import android.app.ActivityOptions;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.graphics.Color;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


JohnThomson:
Rick, what you've done looks reasonable. But a few thoughts:
- I'm not sure why it would look less nice than the old transition, which the issue describes as "come in from the bottom". Is it because the new slide is slower and so more noticeable? Maybe there's an easy way to make it faster?
- Most new features have some way of doing them for older Androids...did you look for an AppCompat or similar approach?
- Issue calls for doing this for bookshelves, too, and I don't see any code for that...

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/72#-:-KyWdQ5aOOnht3loofDu:bhylb2n)*

rjconrad:
-I don't think it's just a different direction, but a different type of transition. The default one does come from the bottom but it's also much more subtle. I did look to see if I could configure it somehow, but I didn't really find anything.

-The documentation is very specific that there is no backwards support for Activity transitions

-That's true about the bookshelves I kind of wanted to get John Hatton's reaction to the look of it before I spent more time on it. 

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/72#-:-KyaAUU_fzzmAacRbnJH:b-8hnocx)*

PR #76, opened 2017-11-13
[BL-5242] Read RTL books from right to left.

rjconrad:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/76)


JohnThomson:
:lgtm:, with a possible optimization to think about.

---

Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 370 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-Kyp1-v_Ow1HwlQdFuvJ:-Kyr8gR9Rd-goVPWQzax:b4drhy1) ([raw file](https://github.com/bloombooks/bloomreader/blob/2f6a2f5f308da7ee03dd3e46d62bf18ce8e53a59/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L370)):*

I followed the pattern of what we're already doing, but it seems to me like we'd be better off using JSoup to parse the html instead of these regex's. (https://jsoup.org/)
</blockquote></details>

You're probably right...one of these days the regular expressions will get annoying or unreliable enough to push a refactoring. I was trying to avoid the time/memory use of making a DOM out of the whole book...maybe jsoup can be used without doing that? But the manipulations are getting more and more complex, and the RE's uglier.
For now, keeping the RE is probably simpler. 

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 373 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-KyrA6NeMVcaQVfyEibG:-KyrA6NeMVcaQVfyEibH:b-rbs8pt) ([raw file](https://github.com/bloombooks/bloomreader/blob/2f6a2f5f308da7ee03dd3e46d62bf18ce8e53a59/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L373)):*
> ```Java
>         Matcher matcher = pattern.matcher(html);
>         return matcher.find();
>     }
> ```

It's unfortunate that in the common case of books that are NOT RTL, we have to search the whole HTML document for this pattern. I'm pretty sure it would reliable to search just the startFrame text that is created a bit further down...it has every thing before the first content page.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 495 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-Kyp1ROvtgP4VHGz2mz9:-KyrAffxSXcsMGdtThJa:b8qgulm) ([raw file](https://github.com/bloombooks/bloomreader/blob/2f6a2f5f308da7ee03dd3e46d62bf18ce8e53a59/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L495)):*

This isn't directly related to the RTL issue - I just happened to notice the error messages because only media books have a WebAppInterface.
</blockquote></details>

Question pages don't have one, even in an MM book, so I think you will find this test has arrived since you wrote this...you will probably have to rebase. Make sure the test doesn't end up in twice.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-:-KyrAw_Is7OwFEiGHhjS:b-ywkc90)*

andrew-polk:
Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 370 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-Kyp1-v_Ow1HwlQdFuvJ:-KyrC__bIsJ6DxDaswNb:b-yzs1t6) ([raw file](https://github.com/bloombooks/bloomreader/blob/2f6a2f5f308da7ee03dd3e46d62bf18ce8e53a59/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L370)):*

You're probably right...one of these days the regular expressions will get annoying or unreliable enough to push a refactoring. I was trying to avoid the time/memory use of making a DOM out of the whole book...maybe jsoup can be used without doing that? But the manipulations are getting more and more complex, and the RE's uglier.
For now, keeping the RE is probably simpler. 
</blockquote></details>

I've suggested we at least consider adding info to meta.json indicating if a book is RTL. Let's not merge until we have that conversation.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76)*

JohnThomson:
:lgtm: But Andrew still wants to have the conversation about putting RTL info in meta.json. So let's wait to merge.

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-:-Kz-WZSctqousnZ7ckWq:b-qelhz4)*

andrew-polk:
We will add this to meta.json per https://issues.bloomlibrary.org/youtrack/issue/BL-5336

andrew-polk:
Now the isRtl property in meta.json can be checked rather than relying on regex (as of 4.0.3187).

rjconrad:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 370 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-Kyp1-v_Ow1HwlQdFuvJ:-L-aqqr9Zpi9BZShcsrE:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/2f6a2f5f308da7ee03dd3e46d62bf18ce8e53a59/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L370)):*

I've suggested we at least consider adding info to meta.json indicating if a book is RTL. Let's not merge until we have that conversation.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76)*

andrew-polk:
Reviewed 2 of 2 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/BloomFileReader.java, line 107 at r3](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-L-asExUtaasA2-_aTZU:-L-asExUtaasA2-_aTZV:b-4neb13) ([raw file](https://github.com/bloombooks/bloomreader/blob/ddc8a2987a59fec1424902f514cc4ec1f8e96605/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java#L107)):*
> ```Java
>     }
> 
>     private JSONObject metaProperties(){
> ```

We now have two different ways for ReaderActivity to get data from meta.json (see brandingProjectName which is read in BookCollection). I would love to see these made consistent.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76)*

rjconrad:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/BloomFileReader.java, line 107 at r3](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-L-asExUtaasA2-_aTZU:-L-at4LElWwDmTTPxRfe:b-5szwku) ([raw file](https://github.com/bloombooks/bloomreader/blob/ddc8a2987a59fec1424902f514cc4ec1f8e96605/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java#L107)):*

We now have two different ways for ReaderActivity to get data from meta.json (see brandingProjectName which is read in BookCollection). I would love to see these made consistent.
</blockquote></details>

Bother - I did look to see if we were using it already, but didn't see that anywhere.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76)*

andrew-polk:
:lgtm:

---

Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/BloomFileReader.java, line 107 at r3](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-L-asExUtaasA2-_aTZU:-L-b7rO-_hTlrv8H1C8q:bi9pw44) ([raw file](https://github.com/bloombooks/bloomreader/blob/ddc8a2987a59fec1424902f514cc4ec1f8e96605/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java#L107)):*

Bother - I did look to see if we were using it already, but didn't see that anywhere.
</blockquote></details>

After discussion, we decided to stay with what we have for now.
The ideal scenario would be to read meta.json once per book into the bookOrShelf object and then have access to that object whenever we need the data. But there is no ideal way to reference class objects between activities. We decided not to change the way ReaderActivity gets the branding information for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/76#-:-L-b8H0T2xPetcSBBXEZ:bnfp4nl)*

PR #91, opened 2018-01-30
[BL-5505] Recover from corrupted FileDescriptor during .bloombundle import

rjconrad:
If a valid FileDescriptor becomes invalid during the extraction, it now restarts the extraction.

For an actual corrupt file (I made one and tested this), the FileDescriptor does not become invalid, and the IOException gets passed up.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/91)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/IOUtilities.java
+++ b/app/src/main/java/org/sil/bloom/reader/IOUtilities.java
@@ -3,6 +3,7 @@ package org.sil.bloom.reader;
 import android.content.Context;
 import android.content.res.AssetManager;
 import android.net.Uri;
+import android.util.Log;
 import android.widget.Toast;
 
 import org.apache.commons.compress.archivers.ArchiveEntry;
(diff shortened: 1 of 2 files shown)


andrew-polk:
Reviewed 2 of 2 files at r1.

---

*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 360 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/91#-L46iMqYv7WsiPkt7DjJ:-L46iMqYv7WsiPkt7DjK:b-cx4rra) ([raw file](https://github.com/bloombooks/bloomreader/blob/b0813d5555af991794f840092300a9675106661f/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L360)):*
> ```Java
>                 // If the IOException is something else, it should go through
>                 if (!fd.valid())
>                     return extractBloomBundle(context, bloomBundleUri);
> ```

What are the chances of us creating an infinite recursion here? I know you mitigated it by only calling in this case, but is this safe enough?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/91)*

rjconrad:

---

*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 360 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/91#-L46iMqYv7WsiPkt7DjJ:-L46j5D0cA7C3r6MzUW-:breq18e) ([raw file](https://github.com/bloombooks/bloomreader/blob/b0813d5555af991794f840092300a9675106661f/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L360)):*

What are the chances of us creating an infinite recursion here? I know you mitigated it by only calling in this case, but is this safe enough?
</blockquote></details>

In order to get an infinite loop we would have to a valid FD that deterministically becomes invalid every single time we run the extraction. I don't see how there could be some bloombundles that don't have that behavior and some that do. So I think it's safe, but with a little more uglifying, I can limit it to 3 attempts.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/91)*

PR #93, opened 2018-01-31
[BL-5505] Recover from corrupted FileDescriptor during .bloombundle import

rjconrad:
Here's the original PR: https://github.com/BloomBooks/BloomReader/pull/91

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/93)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/IOUtilities.java
+++ b/app/src/main/java/org/sil/bloom/reader/IOUtilities.java
@@ -3,6 +3,7 @@ package org.sil.bloom.reader;
 import android.content.Context;
 import android.content.res.AssetManager;
 import android.net.Uri;
+import android.util.Log;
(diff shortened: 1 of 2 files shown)


JohnThomson:
Given the bizarre situation, this seems a reasonable thing to do. If you know anything about how or why file descriptors get corrupted while in use, or what makes it likely that it won't happen repeatedly, that might be worth a comment. Or to say explicitly that we have no idea :-). If it's at all likely it will just go on happening, maybe we should limit the number of retries somehow. Otherwise, LGTM...go ahead and merge.

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/93#-:-L4CMb-wamtkoPOcpaFp:b-8pabwa)*

PR #94, opened 2018-02-01
Bl 5537 Force CQ pages to portrait orientation.

rjconrad:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/94)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -173,7 +173,8 @@
         <activity
             android:name=".ReaderActivity"
             android:label="@string/app_name"
-            android:theme="@style/AppTheme.NoActionBar" />
+            android:theme="@style/AppTheme.NoActionBar"
+            android:configChanges="keyboardHidden|orientation|screenSize|screenLayout"/>
 
         <service
             android:name=".WiFi.NewBookListenerService"
--- a/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
@@ -7,6 +7,7 @@ import android.content.res.Resources;
 import android.graphics.Color;
 import android.graphics.drawable.AnimationDrawable;
 import android.graphics.drawable.ColorDrawable;
+import android.content.res.Configuration;
 import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.Handler;
@@ -82,6 +83,7 @@ public class ReaderActivity extends BaseActivity {
     private int mLastNumberedPageIndex = -1;
     private int mNumberedPageCount = 0;
     private boolean mLastNumberedPageRead = false;
+    private int mOrientation = -1;
     private String mContentLang1 = "unknown";
     int mFirstQuestionPage;
     int mCountQuestionPages;
@@ -122,6 +124,12 @@ public class ReaderActivity extends BaseActivity {
         }
         super.onPause();
         WebAppInterface.stopPlaying();
+    }
+
+    @Override
+    public void onConfigurationChanged(Configuration newConfig){
+        // This prevents the activity from restarting when we change the orientation
+        super.onConfigurationChanged(newConfig);
     }
 
     private void ReportPagesRead()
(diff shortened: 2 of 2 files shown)


JohnThomson:
If I'm reading the constants correctly, we're only controlling the orientation if the user has turned off sensor orientation. So if the sensor orientation is on, and the user has put the device in landscape mode to better read a landscape book, it will stay that way even for CQs. Is that right? Is it what we want? Quite possibly, but maybe worth checking with @hatton.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 131 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/94#-L4HMQTdVQBFYhVyJ8rO:-L4HMQTeEvz7-pMyLVVd:b-gx0jxb) ([raw file](https://github.com/bloombooks/bloomreader/blob/3f8322ce5a4664409d2120501ac7d7c6f9b0e386/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L131)):*
> ```Java
>     @Override
>     public void onConfigurationChanged(Configuration newConfig){
>         // This prevents the activity from restarting when we change the orientation
> ```

That's really bizarre. We override a method, call the base method without changing the arguments or doing anything additional, and it produces a different result? If true, a little more explanation might be helpful.
Do we not want it to restart? Can we switch the layout appropriately without a restart?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/94#-:-L4HNvODps02LkM3IF-0:b-j0072v)*

hatton:
Let's go with how it is and ask the PNG folks what they like. They are living with BR all day every day now, and so I trust their views a lot more than my own.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/94#-:-L4HjOaaO7tgB-OUUx3h:b-qvego4)*

rjconrad:
@JohnThompson No, that's not what those constants mean.  SCREEN_ORIENTATION_SENSOR_LANDSCAPE forces the display to be landscape, but the user can switch between the two possible landscape orientations by turning his phone. Likewise for Portrait, although most phones do not allow an upside-down portrait orientation in my experience.

---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 131 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/94#-L4HMQTdVQBFYhVyJ8rO:-L4KKrthujp_tyOjXEu0:b26ueut) ([raw file](https://github.com/bloombooks/bloomreader/blob/3f8322ce5a4664409d2120501ac7d7c6f9b0e386/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L131)):*

That's really bizarre. We override a method, call the base method without changing the arguments or doing anything additional, and it produces a different result? If true, a little more explanation might be helpful.
Do we not want it to restart? Can we switch the layout appropriately without a restart?
</blockquote></details>

I was going to say that, without implementing this, Android starts the whole activity over with onCreate. Besides gross inefficiency, this was causing the activity to reset the screen orientation to the book's orientation (landscape). Then it was jumping back to the CQ page which set the orientation back to portrait and started the process all over again. It would thrash like that for a few seconds and finally drop me off on page 1.

On second look, however, implementing the override isn't what counts - all we needed was the line in the manifest telling it that this activity should handle config changes. So I've dropped it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/94#-:-L4KIJDL1aZl6o9XFM7A:b-tjsobb)*

andrew-polk:
:lgtm:

---

Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/94#-:-L4Lxa-hRARMDdT5diYx:bnfp4nl)*

PR #95, opened 2018-02-02
BL-5555 Fix for cached images

eomerdws:
Please review the fix below for the issue BL-5555. I am a little concerned that the reload of the web view may cause other issues, but I could not see any. I did this because after clearing the cache it also cleared out bloomPlayer.js. Unfortunately like a web browser there is no way to clear just images.

Note, sorry for the reordering of the imports. I think I still have Android Studio set to automatically clean those. I can fix that if need be.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/95)


eomerdws:
This did not fix it. I just tested it on another device and it failed miserably. Sorry.

PR #96, opened 2018-02-02
Fix for BL-5555 and caching images.

eomerdws:
Sorry the last PR my fork acted really weird and scrubbed the correct code when I committed it and then kept the wrong stuff. Not sure what happened. 

Any rate give this a go. I'm not sure if clearing the cache at this point is the best solution, but it does seem to fix the problem on at least five separate devices. We also could not find any other adverse effects on any of the other features.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/96)


andrew-polk:
Superceded by #97.

PR #100, opened 2018-02-16
[BL-5585] Enable Picture Landscape Mode features 

rjconrad:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/100)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -173,7 +173,8 @@
         <activity
             android:name=".ReaderActivity"
             android:label="@string/app_name"
-            android:theme="@style/AppTheme.NoActionBar" />
+            android:theme="@style/AppTheme.NoActionBar"
+            android:configChanges="keyboardHidden|orientation|screenSize|screenLayout" />
 
         <service
             android:name=".WiFi.NewBookListenerService"
--- a/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java
+++ b/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java
@@ -12,16 +12,13 @@ import org.sil.bloom.reader.models.BookOrShelf;
 import java.io.File;
 import java.io.IOException;
 
-/**
- * Created by rick on 9/26/17.
- */
-
 public class BloomFileReader {
 
     private Context context;
     private String bloomFilePath;
     private Uri bookUri;
     private File bookDirectory;
+    private JSONObject metaProperties;
 
     private static final String CURRENT_BOOK_FOLDER = "currentbook";
     private static final String VALIDATE_BOOK_FILE_FOLDER = "validating";
@@ -98,25 +95,33 @@ public class BloomFileReader {
     }
 
     public boolean getBooleanMetaProperty(String property, boolean defaultIfNotFound){
-        JSONObject properties = metaProperties();
+        JSONObject properties = getMetaProperties();
         if(properties == null)
             return defaultIfNotFound;
         return properties.optBoolean(property, defaultIfNotFound);
     }
 
-    private JSONObject metaProperties(){
-        try {
-            File metaFile = new File(bookDirectory + File.separator + META_JSON_FILE);
-            if (!metaFile.exists())
-                throw new IOException(META_JSON_FILE + " not found");
-            return new JSONObject(IOUtilities.FileToString(metaFile));
-        }
-        catch (JSONException | IOException e){
-            Log.e("BloomFileReader", "Error parsing meta.json: " + e.getMessage());
-            e.printStackTrace();
-            return null;
-        }
+    public String getStringMetaProperty(String property, String defaultIfNotFound){
+        JSONObject properties = getMetaProperties();
+        if(properties == null)
+            return defaultIfNotFound;
+        return properties.optString(property, defaultIfNotFound);
+    }
 
+    private JSONObject getMetaProperties(){
+        if(metaProperties == null) {
+            try {
+                File metaFile = new File(bookDirectory + File.separator + META_JSON_FILE);
+                if (!metaFile.exists())
+                    throw new IOException(META_JSON_FILE + " not found");
+                metaProperties = new JSONObject(IOUtilities.FileToString(metaFile));
+            } catch (JSONException | IOException e) {
+                Log.e("BloomFileReader", "Error parsing meta.json: " + e.getMessage());
+                e.printStackTrace();
+                return null;
+            }
+        }
+        return metaProperties;
     }
 
     private void openFile(String path) throws IOException{
--- a/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
@@ -3,8 +3,10 @@ package org.sil.bloom.reader;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ActivityInfo;
+import android.content.res.Configuration;
 import android.os.AsyncTask;
 import android.os.Bundle;
+import android.support.annotation.Nullable;
 import android.support.v4.view.PagerAdapter;
 import android.support.v4.view.ViewPager;
 import android.util.Log;
@@ -75,19 +77,26 @@ public class ReaderActivity extends BaseActivity {
     private ViewPager mPager;
     private BookPagerAdapter mAdapter;
     private String mBookName ="?";
-    private int mOrientation = ActivityInfo.SCREEN_ORIENTATION_SENSOR_PORTRAIT;
-    private BloomFileReader mFileReader;
     private int mAudioPagesPlayed = 0;
     private int mNonAudioPagesShown = 0;
     private int mLastNumberedPageIndex = -1;
     private int mNumberedPageCount = 0;
     private boolean mLastNumberedPageRead = false;
+    private boolean mInPictureStoryMode = false;
+    private int mOrientation = ActivityInfo.SCREEN_ORIENTATION_UNSPECIFIED;
     private String mContentLang1 = "unknown";
     int mFirstQuestionPage;
     int mCountQuestionPages;
-    WebView mCurrentView;
+    ScaledWebView mCurrentView;
     String[] mBackgroundAudioFiles;
     float[] mBackgroundAudioVolumes;
+
+    enum ViewCtrlState{
+        fixedFullPage,
+        fixedPictureStory,
+        notFixed
+    }
+    private ViewCtrlState mViewCtrl;
 
     // Keeps track of whether we switched pages while audio paused. If so, we don't resume
     // the audio of the previously visible page, but start this page from the beginning.
@@ -99,6 +108,7 @@ public class ReaderActivity extends BaseActivity {
     private boolean mIsMultiMediaBook;
     private boolean mRTLBook;
     private String mBrandingProjectName;
+    private String mFailedToLoadBookMessage;
 
     @Override
     protected void onCreate(Bundle savedInstanceState) {
@@ -126,6 +136,12 @@ public class ReaderActivity extends BaseActivity {
         WebAppInterface.stopAllAudio();
     }
 
+    @Override
+    public void onConfigurationChanged(Configuration newConfig){
+        super.onConfigurationChanged(newConfig);
+        mInPictureStoryMode = determineIfPictureStoryMode(mViewCtrl);
+    }
+
     private void ReportPagesRead()
     {
         try {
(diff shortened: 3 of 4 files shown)


JohnThomson:
I'm confused. I see an optimization of caching metaProperties. I see a new method to get a particular one of them, but nothing calls it. I see new reset() and resetAll() functions, but nothing calls them. I see a bunch of new logging. The only thing that looks as if it MIGHT achieve ac actual change in program behavior is the new android:configChanges. But I don't know what that means OR in what way which Picture Landscape features need to be enabled. BL-5585 is supposed to be about enabling video playback. Can you explain a bit?
Sorry, maybe I'm just not thinking clearly today :-)

---

Reviewed 4 of 4 files at r1.

---

*[app/src/main/java/org/sil/bloom/reader/BloomFileReader.java, line 25 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5Vd0AP2a2Gns4ac2ON:-L5Vd0AP2a2Gns4ac2OO:b29x736) ([raw file](https://github.com/bloombooks/bloomreader/blob/d7af0513ecc5c53931c40e8794c46c23dc63f1a4/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java#L25)):*
> ```Java
>     private Uri bookUri;
>     private File bookDirectory;
>     private JSONObject metaProperties;
> ```

I don't like using the exact same name for the variable and function. Too easy to confuse. Maybe metaPropertiesCache?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L5VeOuUCNL0Kk9fQKYc:belqyl2)*

andrew-polk:
Reviewable kindly thought the diff for  ReaderActivity.java was too big, so it didn't show it to you by default...

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L5Vg0ouO--SBVW5bPh0:bjnf1a8)*

andrew-polk:
*[app/src/main/java/org/sil/bloom/reader/WebAppInterface.java, line 79 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5UWq48L-aAp4gdkLZM:-L5UWq48L-aAp4gdkLZN:b-n115p4) ([raw file](https://github.com/bloombooks/bloomreader/blob/d7af0513ecc5c53931c40e8794c46c23dc63f1a4/app/src/main/java/org/sil/bloom/reader/WebAppInterface.java#L79)):*
> ```Java
>     }
> 
>     // Mainly this is to prevent the Paused state from persisting from one book to another
> ```

Is this a bug we should consider fixing for 1.0? I.e. does it affect playback of audio books?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

rjconrad:

---

*[app/src/main/java/org/sil/bloom/reader/WebAppInterface.java, line 79 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5UWq48L-aAp4gdkLZM:-L5iSZLA5L-2_k5YvyrA:bs472u1) ([raw file](https://github.com/bloombooks/bloomreader/blob/d7af0513ecc5c53931c40e8794c46c23dc63f1a4/app/src/main/java/org/sil/bloom/reader/WebAppInterface.java#L79)):*

Is this a bug we should consider fixing for 1.0? I.e. does it affect playback of audio books?
</blockquote></details>

I believe it does, and for my part I'll say I was quite confused more than once as to why a book wasn't playing, and it turned out that I had paused the last book that I was looking at.  Maybe that doesn't happen so much to real readers though...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

andrew-polk:
*[app/src/main/java/org/sil/bloom/reader/WebAppInterface.java, line 79 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5UWq48L-aAp4gdkLZM:-L5tyVE6sEv8TFSeTrLf:b-mi2369) ([raw file](https://github.com/bloombooks/bloomreader/blob/d7af0513ecc5c53931c40e8794c46c23dc63f1a4/app/src/main/java/org/sil/bloom/reader/WebAppInterface.java#L79)):*

I believe it does, and for my part I'll say I was quite confused more than once as to why a book wasn't playing, and it turned out that I had paused the last book that I was looking at.  Maybe that doesn't happen so much to real readers though...
</blockquote></details>

Rick and I discussed this. He is going to put this specific fix in 1.0.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

rjconrad:
Review status: 3 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/src/main/java/org/sil/bloom/reader/BloomFileReader.java, line 25 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5Vd0AP2a2Gns4ac2ON:-L5x6LwCq8qmBH90grGI:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/d7af0513ecc5c53931c40e8794c46c23dc63f1a4/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java#L25)):*

I don't like using the exact same name for the variable and function. Too easy to confuse. Maybe metaPropertiesCache?
</blockquote></details>

Done.

---

*[app/src/main/java/org/sil/bloom/reader/WebAppInterface.java, line 79 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5UWq48L-aAp4gdkLZM:-L5x6OiWiOheNS8hhYOW:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/d7af0513ecc5c53931c40e8794c46c23dc63f1a4/app/src/main/java/org/sil/bloom/reader/WebAppInterface.java#L79)):*

Rick and I discussed this. He is going to put this specific fix in 1.0.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

andrew-polk:
@johnthomson, have you had a chance to review ReaderActivity.java yet?

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L5xl4Xuw5AqV-Rc6z1O:b-c2n5iv)*

JohnThomson:
I tried but the large chunks of moved (and possibly changed?) code make it very difficult. Hoping Rick can clarify. I'll try to take another look today.

---

Review status: 3 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 174 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L5ucMa3O5RLmllBZ87J:-L5ucMa3O5RLmllBZ87K:b-bagwrq) ([raw file](https://github.com/bloombooks/bloomreader/blob/cba07c850e55f8bad0e84b25b2565e492ade266a/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L174)):*
> ```Java
>     // So we do complicated things with a timer to make sure we don't flip the page too soon...
>     // the minimum delay is specified in MIN_PAGE_SWITCH_MILLIS.
>     public void pageAudioCompleted() {
> ```

I see that a chunk of this is moved rather than deleted, but that makes it really hard to review, since I can't tell whether you also changed things. Where possible, I prefer to use a separate commit for changes that only re-order code. Can you point out anything significant that changed in the moved code?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L5yJrgVXbnwSPHugrQ7:bi3g0cx)*

rjconrad:
Review status: 3 of 4 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 342 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L60tQEoRWi6M2Fqxq6j:-L60tQEoRWi6M2Fqxq6k:b-ts5nmk) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L342)):*
> ```Java
>             this.audioDirectoryPath = audioDirectoryPath;
>         }
> ```

I don't think I changed any of this stuff.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 403 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L60s1AMZUfjJSN4STcN:-L60s1AMZUfjJSN4STcO:b-w635wg) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L403)):*
> ```Java
>     // the minimum delay is specified in MIN_PAGE_SWITCH_MILLIS.
>     public void pageAudioCompleted() {
>         if (!mInPictureStoryMode)
> ```

I changed this line.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 548 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L60sAuHQK_PQiQgT6Qt:-L60sAuHQK_PQiQgT6Qu:b-t468ox) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L548)):*
> ```Java
>                 return false;
>         }
>     }
> ```

I added parseViewCtrlProperty() and determineIfPictureStoryMode()

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

rjconrad:
Review status: 3 of 4 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 492 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L60u7P6wTu2uaGHrjH1:-L60u7P6wTu2uaGHrjH2:b-6bp17l) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L492)):*
> ```Java
>         view.setVisibility(View.VISIBLE);
>         view.startAnimation(anim);
>     }
> ```

I don't think I changed any of the rest of this chunk.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

JohnThomson:
The overall name of the commit is very confusing. I was expecting a mode in which pictures are somehow displayed differently in landscape orientation. Only after I started seeing what seemed unrelated changes to timing of animations did I go back to the issue and see that this commit is really about making landscape mode do something like a continuous video.
Suggested a few name and comment changes. Otherwise, basically looks good.

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 85 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6I44LwTK57gAlPNqoO:-L6I44LwTK57gAlPNqoP:b4m0ek9) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L85)):*
> ```Java
>     private int mNumberedPageCount = 0;
>     private boolean mLastNumberedPageRead = false;
>     private boolean mInPictureStoryMode = false;
> ```

This name did not convey anything to me. Bloom books are often  stories involving, if not actually about, pictures. It's not obvious how that could be a mode rather than a property of the content. If I'm understanding correctly, this mode does a few things
- pages advance automatically after the animation/narration
- the one animated picture is enlarged to fill the screen
- others?
Deserves a comment explaining what linked behaviors it controls. As for a name, how about "mInVideoMode" or "mInContinuousAnimationMode" or "mPlayingFullScreenVideo"?
If there's some external reason to call these picture stories...maybe we use that term in some external context?...it's OK to keep it, but then the comment becomes even more important.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 392 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6I2j9oWos3zbIOMGSZ:-L6I2j9oWos3zbIOMGS_:bb3hz8a) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L392)):*
> ```Java
> 
>     // Minimum time a page must be visible before we automatically switch to the next
>     // (if playing audio...usually this only affects pages with no audio)
> ```

Some of this seems to be out of date...probably already was before you moved it...since I don't think we do automatic page flips currently. Ah, but now we do again...but "if playing audio" isn't the right condition...

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 403 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L60s1AMZUfjJSN4STcN:-L6IRiwVe6iHZQ9CXwtQ:b-49op3s) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L403)):*

I changed this line.
</blockquote></details>

If we were fully following the behavior of the old BloomPlayer, NotFixed books (i.e., picture stories that behave differently in landscape and portrait) would also auto-advance in portrait mode.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 472 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6I6aIPh7yQQVmf_w1L:-L6I6aIPh7yQQVmf_w1M:b-3clrac) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L472)):*
> ```Java
>         }
>         mSwitchedPagesWhilePaused = false;
>         final Animation anim = AnimationUtils.loadAnimation(this, R.anim.grow_and_fade);
> ```

This is probably a pre-existing problem I created. But it would be nice to distinguish the animation of the pause button from the Pan/Zoom animation, especially where mentioned so close together (see comment a few lines up about starting the other animation). Maybe just rename anim to pauseButtonAnimation?

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 526 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6I7fAjo6TfKuaegLgv:-L6I7fAjo6TfKuaegLgw:b-bkiuy8) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L526)):*
> ```Java
> 
>     private ViewCtrlState parseViewCtrlProperty(BloomFileReader fileReader){
>         String property = fileReader.getStringMetaProperty("viewCtrl", "fixedFullPage");
> ```

I don't like the various names here. See above for why pictureStory does not suggest a continuous movie mode to me. viewCtrl also does not suggest metadata that affects how animation is handled.

How about calling the metadata property "animation" or "animationMode" or "panZoomBehavior" with values "normal" (or "default" or "standard" or "bookPageLayout"), "fullPageVideo" (or fullPageContinuousAnimation or fullPage{Continuous?}PanZoom)? Then I think the distinction between fixed/notFixed might be better conveyed by giving both options similar names, but with a modifier. e.g. fullPageVideo{Fixed/LandscapeOnly} (or Always/LandscapeOnly or Forced/Landscape).
Once you pick the best names, let's use them both as the file string values and the variable and enumeration item names.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 615 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6IAruE_hUxT0gq3BTL:-L6IAruE_hUxT0gq3BTM:b-29p5tm) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L615)):*
> ```Java
>         int end = matcher.end(2);
>         String classNames = matcher.group(2);
>         String replacementClass = inPictureStoryMode ? "PictureStoryLandscape" : "Device16x9$1";
> ```

That PictureStoryLandscape class may need changing if you settle on a better name.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 629 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6IQ-53lMa-5Hpf_EPO:-L6IQ-53lMa-5Hpf_EPP:b-psyesk) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L629)):*
> ```Java
>         Double longScale = new Double(longSide / new Double(674));
>         Double shortScale = new Double(shortSide / new Double(378));
>         Double scale = Math.min(longScale, shortScale);
> ```

This modified algorithm seems to imply that we will never be displaying a portrait book in landscape mode or vice versa (without re-arranging the page contents). Is that true? What made it so?

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 918 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6IR0mFdSi1iL8iYe8P:-L6IR0mFdSi1iL8iYe8Q:bv3ddg9) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L918)):*
> ```Java
>         public void reload(boolean landscape){
>             // Reloads only occur with a mViewCtrl of notFixed which means we want to swap between
>             // Device16x9Portrait and PictureStoryLandscape
> ```

Why don't we reload when rotating the device in default mode?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L6I3T97oTGs71-6tb_8:b-k8lsen)*

JohnThomson:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 526 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6I7fAjo6TfKuaegLgv:-L6Iw27XpFmz1-6-DuCs:b-762eha) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L526)):*

Yes, naming is hard... nothing would be worse for me than calling anything that isn't video, "video". Similarly, nothing about automatic page-turning says "animation" to me, and it's very furture-fagile. What happens when each page has some game animation, but you don't want to automatically turn the pages? I'm don't see any misunderstandings with the "continuous" word, though it's rather blah. AutoPageAdvance? Or is there more to it then the page advancing? 
</blockquote></details>

As I understand it when this variable is true two things happen: the first picture on the page (or possibly the first that has pan/zoom) is made to fill the screen, and page advance is automatic when the pan/zoom (and narration) is complete or after three seconds. Not quite sure what will happen if the page has narration but no pictures (or a picture but no pan/zoom). fullPagePictureAutoAdvance might work. We could also separate them...fullPagePicture could be one boolean and autoAdvance another...even if for the moment they always have the same value. That isn't necessarily true...the old BloomPlayer had AutoAdvance even in Portrait mode, where it did't mess with the page layout.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100)*

hatton:
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 526 at r2](https://beta.reviewable.io/reviews/bloombooks/bloomreader/100#-L6I7fAjo6TfKuaegLgv:-L6J22-w1OSlSWCUS8VT:b-yui4hf) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L526)):*

As I understand it when this variable is true two things happen: the first picture on the page (or possibly the first that has pan/zoom) is made to fill the screen, and page advance is automatic when the pan/zoom (and narration) is complete or after three seconds. Not quite sure what will happen if the page has narration but no pictures (or a picture but no pan/zoom). fullPagePictureAutoAdvance might work. We could also separate them...fullPagePicture could be one boolean and autoAdvance another...even if for the moment they always have the same value. That isn't necessarily true...the old BloomPlayer had AutoAdvance even in Portrait mode, where it did't mess with the page layout.
</blockquote></details>

So it does seem right for Bloom to put an autoAdvance variable in meta.json, and decouple that from this full page thing. BTW, the full page thing... I suspect we should be implementing that as either a different layout or perhaps a CSS-based toggle on the layout. The more of this stuff that lives in BR instead of html/JS/CSS, the harder it is going to be for us to provide a preview in Bloom.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bloombooks/bloomreader/100)*

rjconrad:
The name PictureStoryLandscape came from CSS class which hides the text and makes the picture full screen. PictureStoryMode is when we're using that class as opposed to Device16x9Portrait or Device16x9Landscape. So I took it from a name that's already in the codebase.

---

Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 403 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L60s1AMZUfjJSN4STcN:-L6LIeQB_CD-yJxNnGqk:br7jrhe) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L403)):*

If we were fully following the behavior of the old BloomPlayer, NotFixed books (i.e., picture stories that behave differently in landscape and portrait) would also auto-advance in portrait mode.
</blockquote></details>

The design I was given was that books should autoadvance in PictureStoryLandscape, but not in portrait.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 526 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6I7fAjo6TfKuaegLgv:-L6LOVWetJe7cLbWf9Bq:b-z3h5pp) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L526)):*

So it does seem right for Bloom to put an autoAdvance variable in meta.json, and decouple that from this full page thing. BTW, the full page thing... I suspect we should be implementing that as either a different layout or perhaps a CSS-based toggle on the layout. The more of this stuff that lives in BR instead of html/JS/CSS, the harder it is going to be for us to provide a preview in Bloom.
</blockquote></details>

It is a CSS class. All BR does is apply that class.

The design I was given was to support 3 types of books:
1 - Regular books 
2 - Books that are always in PictureStoryLandscape and autoadvance
3 - Books that swap based on the device orientation. In a portrait orientation, PictureStoryLandscape is not used so the text is displayed, and the book does not autoadvance. In a landscape orientation, PictureStoryLandscape is used, hiding the words, and the book autoadvances.

Therefore viewCtrl toggles between these three options. So yeah, it wasn't a very easy variable to name.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 629 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6IQ-53lMa-5Hpf_EPO:-L6LL4-Jz9mP8dFtKYhS:bleg5mu) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L629)):*

This modified algorithm seems to imply that we will never be displaying a portrait book in landscape mode or vice versa (without re-arranging the page contents). Is that true? What made it so?
</blockquote></details>

That's right. That was the design I was given back when I first implemented landscape books. Landscape books are always displayed in a landscape orientation, and portrait books in a portrait orientation. See getPageOrientationAndRotateScreen() which sets the book orientation based on the css class of the page element.

---

*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 918 at r2](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-L6IR0mFdSi1iL8iYe8P:-L6LLyjdFcyyu482zswA:be10ymr) ([raw file](https://github.com/bloombooks/bloomreader/blob/cf51afd8bc8b9dc5fc58d76a2e3644b5207b1ec1/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L918)):*

Why don't we reload when rotating the device in default mode?
</blockquote></details>

This method is called by onSizeChanged() when we've switched in or out of PictureStoryMode. The whole purpose of this method is to update the html for that change and reload it. There's no other scenario right now in which a  webview needs to edit the html and reload.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L6LJOY9BdqjYxSILSAp:b-l4edrd)*

JohnThomson:
That probably means that PictureStoryLandscape came from me in the first place! And here I am complaining about it.
It's an unfortunate fact of programming life that once the client sees what they asked for, they sometimes realize that isn't quite what they wanted. I don't think I've actually seen the spec you're working from...it seems to have more detail than the YouTrack issue. But at this point, JohnH and I are both thinking that it might make things both clearer and more flexible to control the landscape=>full screen picture behavior independently from the auto-advance behavior. And that's leading to further ideas. We may need you to redo some bits of this once we get our thinking clearer. Sorry. It's not your fault. At this point I think it's best to wait until we clarify what we want before you change anything.

---

Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L6ORb8W52Vb04ZwoFVV:b-fgewf3)*

hatton:
> That probably means that PictureStoryLandscape came from me in the first place! And here I am complaining about it.

I know you're just being humble, but I want to assert that it is MORE than OK for any of us to decide we don't like any of our own previous decision. On the contrary, we should applaud it.

---

Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/100#-:-L6OrkZjurePhEDw2P3n:bhmgpa6)*

rjconrad:
So perhaps...

Two meta-properties:
 - autoAdvance with the values off, on, or pictureOnlyLandscape
 - pictureOnlyLandscape with the values off, on, or deviceOrientation

Setting pictureOnlyLandscape to deviceOrientation would enable pictureOnlyLandscape mode when the device is in a landscape orientation. Setting autoAdvacne to pictureOnlyLandscape would enable auto-advance when in pictureOnlyLandscape mode.

Rename the css class to PictureOnlyLandscape and the Java variables to line up with the meta-property names.

JohnThomson:
Two meta properties would be one option. John and I were also wondering whether the picture only landscape thing ties in with the page-size-and-orientation setting that Bloom already has...there are already options Device16x9{Portrait/Landscape}, maybe a third option there might control the pictureOnlyLandscape behavior? But then a question arises whether this only works for certain page layouts...it gets complicated, I'm not sure we reached a complete conclusion. @hatton, did you write something up? Or was I supposed to?

PR #113, opened 2018-03-09
[BL-5726] Fix path error in thumbnail cleanup

rjconrad:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/113)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java
+++ b/app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java
@@ -18,7 +18,7 @@ import java.io.File;
 
 public class ThumbnailCleanup extends AsyncTask<File, Void, Void> {
 
-    public static final String DID_THUMB_CLEANUP = "didThumbCleanup";
+    public static final String DID_THUMB_CLEANUP = "didThumbCleanupRight";
 
     private SharedPreferences values;
 
@@ -34,10 +34,10 @@ public class ThumbnailCleanup extends AsyncTask<File, Void, Void> {
             return null;
 
         File thumbsDirectory = files[0];
-        String[] paths = thumbsDirectory.list();
-        for(String path : paths){
-            if (path.endsWith(".png"))
-                new File(path).delete();
+        String[] filenames = thumbsDirectory.list();
+        for(String filename : filenames){
+            if (filename.endsWith(".png"))
+                new File(thumbsDirectory + File.separator + filename).delete();
         }
 
         SharedPreferences.Editor valuesEditor = values.edit();


hatton:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java, line 37 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/113#-L7A9pS_t41gTzYQk2y2:-L7A9pS_t41gTzYQk2y3:bt92yes) ([raw file](https://github.com/bloombooks/bloomreader/blob/71aa4fdedd7500d33e813b66d71983d5612c75d5/app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java#L37)):*
> ```Java
> 
>         File thumbsDirectory = files[0];
>         String[] paths = thumbsDirectory.list();
> ```

It really helps to keep things clear if we use "filenames" and "paths" and never interchangeably. Even lots of APIs I see use them interchangeably, to my chagrin. Possibly it was the choice of "paths" in the original code, when in fact they were not "paths", that led to this bug in the first place. They were "filenames". Happily, you can improve that now :-)

---

*[app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java, line 40 at r1](https://reviewable.io:443/reviews/bloombooks/bloomreader/113#-L7AAwbzavK20Js2FXq_:-L7AAwbzavK20Js2FXqa:b-6pnh5f) ([raw file](https://github.com/bloombooks/bloomreader/blob/71aa4fdedd7500d33e813b66d71983d5612c75d5/app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java#L40)):*
> ```Java
>         for(String path : paths){
>             if (path.endsWith(".png"))
>                 new File(thumbsDirectory + File.separator + path).delete();
> ```

(for my own edification and any other reviewers) Sorry, Paths.get(thumbsDirectory, path); requires API level 26! Oh, JAVA...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/bloombooks/bloomreader/113)*

rjconrad:
Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java, line 37 at r1](https://beta.reviewable.io/reviews/bloombooks/bloomreader/113#-L7A9pS_t41gTzYQk2y2:-L7AEFlc7y6OScfrahK0:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/71aa4fdedd7500d33e813b66d71983d5612c75d5/app/src/main/java/org/sil/bloom/reader/ThumbnailCleanup.java#L37)):*

It really helps to keep things clear if we use "filenames" and "paths" and never interchangeably. Even lots of APIs I see use them interchangeably, to my chagrin. Possibly it was the choice of "paths" in the original code, when in fact they were not "paths", that led to this bug in the first place. They were "filenames". Happily, you can improve that now :-)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/bloombooks/bloomreader/113)*

PR #126, opened 2018-05-08
[BL-5964] Improved bloombundle import.

rjconrad:
Better handle the mysterious filedescriptor-goes-bad problem by not starting over completely. 
Provide feedback during import. 
Show a helpful message if we run out of space.

For https://issues.bloomlibrary.org/youtrack/issue/BL-5964

The two problems that I identified are that importing a large bundle took an excessive amount of time with no user feedback, making the app appear frozen, and that there was no error message for running out of space. I fixed the first by making the import a background job with periodic feedback, and reduced the time needed for import by recovering from an expired filedescriptor at the file that failed rather than at the beginning. If you do run out of space, you will get to keep the books we were able to copy.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/126)


andrew-polk:
Reviewed 4 of 4 files at r1.

---

*[app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java, line 108 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/126#-LC-qXiGAyrfuDDHp6GF:-LC-qXiGAyrfuDDHp6GG:b-s26h4m) ([raw file](https://github.com/bloombooks/bloomreader/blob/5c91b14558de8b1ba581c620b0bd2ffb29b3167e/app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java#L108)):*
> ```Java
>                 if (!fd.valid()) {
>                     fd = mainActivity.getContentResolver().openFileDescriptor(bloomBundleUri, "r").getFileDescriptor();
>                     tarInput = new TarArchiveInputStream(new FileInputStream(fd));
> ```

Do we need to close the existing tarInput?
Shouldn't we close this one?

---

*[app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java, line 122 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/126#-LC-pzt78m_97S1pgWNe:-LC-pzt78m_97S1pgWNf:bc976xt) ([raw file](https://github.com/bloombooks/bloomreader/blob/5c91b14558de8b1ba581c620b0bd2ffb29b3167e/app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java#L122)):*
> ```Java
>     }
> 
>     private static String extractTarEntry(TarArchiveInputStream tarInput, String targetPath) throws IOException {
> ```

Is there any reason this method can't stay in IOUtilities? If it is generic enough to be used by other processes in the future, it feels like it belongs there.

Is it because tarInput is no longer being closed on the success path?

---


*Comments from [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/126)*

rjconrad:
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions.

---

*[app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java, line 108 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/126#-LC-qXiGAyrfuDDHp6GF:-LC2o8yYEcuI02ApyxFL:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/5c91b14558de8b1ba581c620b0bd2ffb29b3167e/app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java#L108)):*

Do we need to close the existing tarInput?
Shouldn't we close this one?
</blockquote></details>

Done.

---

*[app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java, line 122 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/126#-LC-pzt78m_97S1pgWNe:-LC2oCMQ8_WgwZZXx43f:b-ta2cpe) ([raw file](https://github.com/bloombooks/bloomreader/blob/5c91b14558de8b1ba581c620b0bd2ffb29b3167e/app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java#L122)):*

Is there any reason this method can't stay in IOUtilities? If it is generic enough to be used by other processes in the future, it feels like it belongs there.

Is it because tarInput is no longer being closed on the success path?
</blockquote></details>

It's no longer written quite the way you would write a generic tar extracter because this method only extracts one entry. If we do need a generic one later though, we could write one to use this method, so it's probably a good idea to have it in IOUtilities.
Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/126)*

rjconrad:

---

*[app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java, line 108 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/126#-LC-qXiGAyrfuDDHp6GF:-LC3qiUm4AFB6cYy21v0:b-gv9ex7) ([raw file](https://github.com/bloombooks/bloomreader/blob/5c91b14558de8b1ba581c620b0bd2ffb29b3167e/app/src/main/java/org/sil/bloom/reader/ImportBundleTask.java#L108)):*

I think we're still not closing this new tarInput.
</blockquote></details>

The while(true) is going to carry us back into the try block and the new tarInput will get closed in line 98 right before the return.

---


*Comments from [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/126)*

andrew-polk:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/126#-:-LC3rxgX1JHb0W5ULpMj:bnfp4nl)*

PR #139, opened 2018-10-03
[BL-5969] Rework intent filters for receiving bloom files.

rjconrad:
Issue: https://issues.bloomlibrary.org/youtrack/issue/BL-5969

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/139)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -43,136 +43,74 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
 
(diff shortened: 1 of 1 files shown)


hatton:
I'm finding this very hard to review. It removes huge chunks of descriptions that were there for a purpose, and contained extensive comments (also removed). Presumably because you have decided none of it is needed. And the new stuff... no comments at all. So could well be correct, but basically impossible to review.

rjconrad:
> Do I have that right?

No. That's not what I meant, and that's not what I said.

PR #167, opened 2019-03-26
Toggle video pause/play on actual playing or not (BL-7003)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/167)


andrew-polk:
The other thing to consider with regard to changes to BloomReader is that the more complex they are, the more it might be worth putting it off to the next version (the React Native one).

PR #168, opened 2019-03-28
Make importing big books more reliable on old devices (BL-6970)

StephenMcConnel:
It's my hypothesis that older devices are more susceptible to the
problem addressed here.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/168)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/IOUtilities.java
+++ b/app/src/main/java/org/sil/bloom/reader/IOUtilities.java
@@ -5,6 +5,7 @@ import android.content.res.AssetManager;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Environment;
+import android.util.Log;
 import android.widget.Toast;
 
 import org.apache.commons.compress.archivers.ArchiveEntry;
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/sil/bloom/reader/IOUtilities.java
```diff
@@ -164,20 +165,30 @@ public static boolean copyAsset(AssetManager assetManager, String fromAssetPath,
     }
 
     public static boolean copyFile(InputStream fromStream, String toPath) {
+        int totalRead = 0;
         try {
             new File(toPath).createNewFile(); //this does nothing if if already exists
             OutputStream out = new FileOutputStream(toPath);
-            byte[] buffer = new byte[1024];
+            // Reading the file 1024 bytes at a time runs into trouble on some older
```
hatton: Why would that be?

StephenMcConnel:
___
*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 172 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/168#-Lb5biq1nUdWPrnr0h8C-r1-172:-Lb8zU4Y0dqhBXZMtSh0:bx7bidx) ([raw file](https://github.com/bloombooks/bloomreader/blob/168c30392305aca52ba4c03be309de091031c466/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L172)):*

Why would that be?
</blockquote></details>

My hypothesis is that another thread in Bloom Reader tries to open the new book it finds in the Bloom folder before the copy finishes, closing the output stream prematurely.  Having a larger buffer makes the copy go faster so that it finishes earlier.  I don't know enough about Bloom Reader internals to know whether this is really what is happening, but it's the only plausible scenario I've thought of.  I could continue researching this, but I think the current two-fold fix is still worthwhile.

PR #175, opened 2019-04-11
Remove unreliable file descriptor operations (BL-7073)

StephenMcConnel:
See https://stackoverflow.com/questions/20080171/contentresolver-openfiledescriptor-does-not-call-contentprovider-openfile
for why these changes are needed and appropriate.  A similar change has
already been made to IOUtilities.copyBloomdFile(context, uri, toPath).
(This method used to be called copyFile, but is used only for copying
.bloomd files, so now has an added check in it.)

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/175)


andrew-polk:
What issue is this related to?

andrew-polk:
Found it. BL-7073
Let's get that in the commit comment, please.

PR #186, opened 2019-06-14
BL-7256 Include some book features in analytics report

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/186)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java
+++ b/app/src/main/java/org/sil/bloom/reader/BloomFileReader.java
@@ -5,6 +5,7 @@ import android.net.Uri;
 import android.support.annotation.Nullable;
 import android.util.Log;
 
+import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.sil.bloom.reader.models.BookCollection;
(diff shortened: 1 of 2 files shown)


andrew-polk:
___
*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 314 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/186#-LhaEpLq4BjIyF3636TU:-LhaGMHM-2U9whphY10c:b-1fqwj9) ([raw file](https://github.com/bloombooks/bloomreader/blob/11bf39ceed962d4fa6676d41cc0f5cdae92cfa86/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L314)):*

> ```
> (hasImageDescriptions ? "blind," : "")
> ```
I'm a bit unsure of this definition for the "blind" feature. Consider a book with images, no image descriptions, but all the text is narrated. And consider the extreme case of a book with no images but all the text is narrated. We may not want to change the definition just for that extreme case, but what about the first one?
</blockquote></details>

Ok, so I guess we already have talkingBook as another feature, so maybe this is the best we can do.

PR #221, opened 2019-11-15
Run BookCollection.init() asynchronously and add progress bar (BL-7432)

StephenMcConnel:
Users may still have to wait for all the books to be loaded, but there's
some sign of work being done rather than a possibly frozen device.

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/221)


hatton:
Merging based on Andrew's code approval above and my UI approval.

PR #254, opened 2020-09-22
Don't crash if the device doesn't have GPS

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/254)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/MainActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/MainActivity.java
@@ -209,7 +209,19 @@ public class MainActivity extends BaseActivity
         // getting them, so don't bother with a minimum distance.
         long minDistanceM = 0;
         try {
-            lm.requestLocationUpdates(LocationManager.GPS_PROVIDER, minTimeMs, minDistanceM, new LocationListener() {
+            List<String> allProviders = lm.getAllProviders();
+            String provider = null;
+            if (allProviders.contains(LocationManager.GPS_PROVIDER))
+                provider = LocationManager.GPS_PROVIDER;
+            else if (allProviders.contains(LocationManager.NETWORK_PROVIDER))
+                provider = LocationManager.NETWORK_PROVIDER;
+            else if (allProviders.contains(LocationManager.PASSIVE_PROVIDER))
+                provider = LocationManager.PASSIVE_PROVIDER;
+            if (provider == null) {
+                Log.e("locationError", "no location provider available");
+                return;
+            }
+            lm.requestLocationUpdates(provider, minTimeMs, minDistanceM, new LocationListener() {
                 // We don't actually want the information, we just want lastKnownLocation to be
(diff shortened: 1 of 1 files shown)


JohnThomson (approved):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/254)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 220 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/254#-MHsUAyo0rc_iEM7VLAx:-MHsUAyo0rc_iEM7VLAy:bccfk7z) ([raw file](https://github.com/bloombooks/bloomreader/blob/bc01006fa735ca207fdb1138a4ce610067a8ce2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L220)):*
> ```Java
>             else if (allProviders.contains(LocationManager.PASSIVE_PROVIDER))
>                 provider = LocationManager.PASSIVE_PROVIDER;
>             if (provider == null) {
> ```

Are there other possibilities? Maybe if we don't have one yet we could just see whether allProviders contains anything, and if so, use the first one? Or, we could try getBestProvider (), maybe with arguments requesting the most accurate? (I started with that, but sometimes Android thinks the network provider is MORE accurate than the GPS one, and we weren't sure we wanted to believe it in a remote location.) But you might consider simply asking for the best (most accurate) provider at once if there is no GPS provider.
However, you said this was urgent, and what you have is already a fine improvement, so I'm going to merge.

JohnThomson (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/254)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

andrew-polk:
___
*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 220 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/254#-MHsUAyo0rc_iEM7VLAx:-MHwG23m9wkgs0cfJZIz:b4i6ma9) ([raw file](https://github.com/bloombooks/bloomreader/blob/bc01006fa735ca207fdb1138a4ce610067a8ce2c/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L220)):*

Are there other possibilities? Maybe if we don't have one yet we could just see whether allProviders contains anything, and if so, use the first one? Or, we could try getBestProvider (), maybe with arguments requesting the most accurate? (I started with that, but sometimes Android thinks the network provider is MORE accurate than the GPS one, and we weren't sure we wanted to believe it in a remote location.) But you might consider simply asking for the best (most accurate) provider at once if there is no GPS provider.
However, you said this was urgent, and what you have is already a fine improvement, so I'm going to merge.
</blockquote></details>

FWIW, those are the only relevant constants on LocationManager.

PR #256, opened 2021-03-15
BL-8993 Enable toggling image descriptions

gmartin7:
* depends on new version of BP

---
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/256)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java
@@ -107,8 +107,9 @@ public class ReaderActivity extends BaseActivity {
             // shipped with this program, combined with a param pointing to the book we just
             // decompressed.
             final String url = "file:///android_asset/bloom-player/bloomplayer.htm?url=file:///"
-                    + bookHtmlFile.getAbsolutePath()+"&showBackButton=true&allowToggleAppBar=true&initiallyShowAppBar=false"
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/256)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[app/src/main/java/org/sil/bloom/reader/ReaderActivity.java, line 112 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/256#-MVrkY0VCD0owi1Nl1Cd:-MVrkY0VCD0owi1Nl1Ce:b-gy06ah) ([raw file](https://github.com/bloombooks/bloomreader/blob/b3572b3989c4d48bb9f78acc6117f77e15e3d976/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L112)):*
> ```Java
>                     + bookHtmlFile.getAbsolutePath()+"&showBackButton=true&allowToggleAppBar=true"
>                     + "&initiallyShowAppBar=false&centerVertically=true&hideFullScreenButton=true"
>                     + "&initiallyReadImageDescriptions=true&independent=false&host=bloomreader";
> ```

I don't understand why we are adding this. `true` is the default value, right?
And actually, we don't need the url param at all in bloom-player, as far as I can tell.

PR #259, opened 2021-06-16
Fix file delete on Android 10+ (BL-9637)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/259)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java
+++ b/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java
@@ -51,8 +51,12 @@ public class FileCleanupTask extends AsyncTask<Uri, Void, Void> {
             if (nonRemovableStorageDir == null)
                 return;
 
+            String fileNameFromUri = fileNameFromUri(uriToCleanUp);
+            if (fileNameFromUri == null || fileNameFromUri.isEmpty())
+                return;
+
             // Returns null if the file is not found
-            File fileToDelete = searchForFile(nonRemovableStorageDir, uriToCleanUp);
+            File fileToDelete = searchForFile(nonRemovableStorageDir, fileNameFromUri);
 
             if (fileToDelete != null)
                 fileToDelete.delete();
@@ -80,10 +84,10 @@ public class FileCleanupTask extends AsyncTask<Uri, Void, Void> {
         return file.getPath().startsWith(getBloomDirectory().getPath());
     }
 
-    private boolean shouldSearchThisDirectory(File dir, Uri uriToCleanUp) throws ExtStorageUnavailableException {
+    private boolean shouldSearchThisDirectory(File dir, String fileName) throws ExtStorageUnavailableException {
         // We don't want to find the bloomd's in our library
         // Bundles are fair game everywhere
-        if (uriToCleanUp.getPath().endsWith(IOUtilities.BLOOM_BUNDLE_FILE_EXTENSION))
+        if (fileName.endsWith(IOUtilities.BLOOM_BUNDLE_FILE_EXTENSION))
             return true;
 
         return !dir.equals(getBloomDirectory());
@@ -95,7 +99,7 @@ public class FileCleanupTask extends AsyncTask<Uri, Void, Void> {
         return bloomDirectory;
     }
 
-    private File searchForFile(File dir, Uri uriToCleanUp) throws ExtStorageUnavailableException {
+    private File searchForFile(File dir, String fileName) throws ExtStorageUnavailableException {
         if (dir == null)
             return null;
 
@@ -104,10 +108,10 @@ public class FileCleanupTask extends AsyncTask<Uri, Void, Void> {
             return null;
 
         for (File f : list) {
-            if (f.isFile() && matchesSearchFile(f, uriToCleanUp))
+            if (f.isFile() && f.getName().equals(fileName))
                 return f;
-            if (f.isDirectory() && shouldSearchThisDirectory(f, uriToCleanUp)) {
-                File fileToDelete = searchForFile(f, uriToCleanUp);
+            if (f.isDirectory() && shouldSearchThisDirectory(f, fileName)) {
+                File fileToDelete = searchForFile(f, fileName);
                 if (fileToDelete != null)
                     return fileToDelete;
             }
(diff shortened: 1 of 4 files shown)


JohnThomson (changes requested):
Looks basically good...a few questions to think about.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/259#-:-McpE1IO3mGsQa9mQ6CD:btu0q7o)* status: all files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java, line 131 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpD0N96fyj92q0ZxJP:-McpD0N96fyj92q0ZxJQ:b-6q9qub) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java#L131)):*
> ```Java
>             String fileNameOrPath = IOUtilities.getFileNameOrPathFromUri(context, contentURI);
>             if (fileNameOrPath != null && !fileNameOrPath.isEmpty())
>                 fileName = IOUtilities.getFilename(fileNameOrPath);
> ```

How likely is it that we are removing useful path information here? For example, perhaps the path is complete enough to let us find the file without searching the whole storage? Or to prevent us deleting, say, a backup of an earlier version of the .bloomd instead of the thing we just downloaded that the URI really points to?
___
*[app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java, line 138 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpBU02Cj9IjXg3bDuc:-McpBU02Cj9IjXg3bDud:b-v0lp8f) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java#L138)):*
> ```Java
>             return fileName;
> 
>         // From here to the end was the original code, but it seems to have stopped working in Android 10.
> ```

Would the logic here be useful to the other caller of getFileNameOrPathFromUri? If so, maybe we should move it into the method?
___
*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 501 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpDn4v0tuVH9Y3YFqT:-McpDn4wEXtKGrmQPOSL:bjngc4h) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L501)):*
> ```Java
>     }
> 
>     static String getFileNameOrPathFromUri(Context context, Uri uri) {
> ```

Is it at all feasible to unit test this? Since I see you wrote unit tests for another method of this class (no good deed goes unpunished :-)
___
*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 504 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpB-658PbRvGII7aGR:-McpB-658PbRvGII7aGS:b-5oa3sq) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L504)):*
> ```Java
>         String nameOrPath = uri.getPath();
>         // Content URI's do not use the actual filename in the "path"
>         if (uri.getScheme().equals("content")) {
> ```

How sure are you that the only URIs we can be passed (that correspond to a file we would want to delete) are content URIs?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/259)* status: 2 of 4 files reviewed, 4 unresolved discussions (waiting on @JohnThomson)

___
*[app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java, line 131 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpD0N96fyj92q0ZxJP:-MdIs2dSCCJAdsb51B0I:b-adhcmp) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java#L131)):*

How likely is it that we are removing useful path information here? For example, perhaps the path is complete enough to let us find the file without searching the whole storage? Or to prevent us deleting, say, a backup of an earlier version of the .bloomd instead of the thing we just downloaded that the URI really points to?
</blockquote></details>

This is a very good point, and would be a good improvement if we can make it work.
In practice, I haven't been able find anything which gives more than just file name through experimentation. So so far I haven't been able to figure out how I would write this and test it.
We had discussed having the code check that the thing we are deleting ends in whatever we determined from the uri. But that actually opens up the possibility of matching on partial file names which seems less safe.
Of course, if I could figure out what a "path" would actually look like as opposed to just a file name, I might be able to do some kind of ends with where we ensure each directory and the filename are complete matches. But again, since I so far haven't been able to test/experiment, that doesn't seem feasible.

While not ideal, I'm voting for just leaving this as is for now. It is no worse than what we had.
___
*[app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java, line 138 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpBU02Cj9IjXg3bDuc:-MdIV92I9yUcUDPgEwH0:b7a5952) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java#L138)):*

Would the logic here be useful to the other caller of getFileNameOrPathFromUri? If so, maybe we should move it into the method?
</blockquote></details>

I admit it is not ideal, but I'm trying to make minimal safe changes, and this logic was not present for the other caller before.
I've added a comment.
___
*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 501 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpDn4v0tuVH9Y3YFqT:-MdIQJxBBniqSlZrdXaS:b-93vyx8) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L501)):*

Is it at all feasible to unit test this? Since I see you wrote unit tests for another method of this class (no good deed goes unpunished :-)
</blockquote></details>

I looked into this a bit and decided it isn't worth it.
___
*[app/src/main/java/org/sil/bloom/reader/IOUtilities.java, line 504 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpB-658PbRvGII7aGR:-MdIsvwO9znoQ-4mGkZE:bcn31yf) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/IOUtilities.java#L504)):*

How sure are you that the only URIs we can be passed (that correspond to a file we would want to delete) are content URIs?
</blockquote></details>

Definitely not guaranteed, but it is what we expect...
And if we got a file uri, that would have been handled before we got here with an early out in cleanUpUri.

JohnThomson (approved):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/259)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java, line 131 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/259#-McpD0N96fyj92q0ZxJP:-MdJNVYc7FyvyDI60KSU:bakacat) ([raw file](https://github.com/bloombooks/bloomreader/blob/e7ca6c879b4b6b7c7e295d364791d6976d0b65aa/app/src/main/java/org/sil/bloom/reader/FileCleanupTask.java#L131)):*

This is a very good point, and would be a good improvement if we can make it work.
In practice, I haven't been able find anything which gives more than just file name through experimentation. So so far I haven't been able to figure out how I would write this and test it.
We had discussed having the code check that the thing we are deleting ends in whatever we determined from the uri. But that actually opens up the possibility of matching on partial file names which seems less safe.
Of course, if I could figure out what a "path" would actually look like as opposed to just a file name, I might be able to do some kind of ends with where we ensure each directory and the filename are complete matches. But again, since I so far haven't been able to test/experiment, that doesn't seem feasible.

While not ideal, I'm voting for just leaving this as is for now. It is no worse than what we had.
</blockquote></details>

Agreed. If the paths we are actually getting from getFileNameOrPathFromUri() don't contain anything more useful than a filename, then this is the best I can think to do.

PR #264, opened 2021-08-10
Handle new intent without file permissions (BL-10230)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/264)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/MainActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/MainActivity.java
@@ -336,8 +336,17 @@ public class MainActivity extends BaseActivity
     protected void onNewIntent(Intent intent) {
         super.onNewIntent(intent);
         checkForPendingReadBookAnalyticsEvent();
-        alreadyOpenedFileFromIntent = false;
-        processIntentData(intent);
+
+        // It is possible that the user has never given permissions (or revoked them).
+        // In that case, we don't try to handle the intent.
+        // OnResume will get called, and we will just load the main screen (which asks the user for permissions).
+        // Unfortunately, after giving permissions, the user will need to try to open the book
+        // from the file again, but the main goal here is to prevent a crash, and I don't think
+        // it is worth trying to get more fancy (like remembering the intent until he gives permissions).
+        if (haveStoragePermission(this)) {
+            alreadyOpenedFileFromIntent = false;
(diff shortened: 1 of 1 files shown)


JohnThomson (changes requested):
LGTM with one comment improvement.

> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/264#-:-MglPBV3DxVlzctsEPT3:b3duuna)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 341 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/264#-MglOnbB7sak_gregIpQ:-MglOnbB7sak_gregIpR:b-pd2cw7) ([raw file](https://github.com/bloombooks/bloomreader/blob/1451bc8b255aacf5e5c344e97e550800ba63439f/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L341)):*
> ```Java
> 
>         // It is possible that the user has never given permissions (or revoked them).
>         // In that case, we just take them back to the main screen which asks him for permissions.
> ```

I think this needs a little more. It's not obvious how this is accomplished. (I think the answer is, receiving onNewIntent implies the activity stack has already been popped back to this, so unless we do something this MainActivity will resume.)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/264)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[app/src/main/java/org/sil/bloom/reader/MainActivity.java, line 341 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/264#-MglOnbB7sak_gregIpQ:-MglQsRh1dOplGQFJxLa:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/1451bc8b255aacf5e5c344e97e550800ba63439f/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L341)):*

I think this needs a little more. It's not obvious how this is accomplished. (I think the answer is, receiving onNewIntent implies the activity stack has already been popped back to this, so unless we do something this MainActivity will resume.)
</blockquote></details>

Done.

PR #284, opened 2022-05-18
Warn the user if an old WebView is installed (BL-11176)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/284)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -2,42 +2,29 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="org.sil.bloom.reader"
     android:installLocation="auto">
-
-    <!-- Normal Permissions - granted by being in Manifest -->
-    <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
     <!-- So that just for the duration of WiFi transfer we can hold a lock. -->
-    <uses-permission android:name="android.permission.WAKE_LOCK"/>
-    <!-- Used at least for analytics, soon for Bloom Library download. Was present when I added
-    analytics, so may be used for something else also. -->
-    <uses-permission android:name="android.permission.INTERNET" />
-    <!-- On later versions of Android this seems to be needed for receiving UDP broadcasts. -->
-    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
-    <!-- On later versions of Android this seems to be needed for receiving UDP broadcasts. -->
-    <uses-permission android:name="android.permission.CHANGE_WIFI_MULTICAST_STATE" />
-    <!-- currently just because it's how we get the device name for WiFi, but eventually we will have
-    other Bluetooth functions. -->
-    <uses-permission android:name="android.permission.BLUETOOTH"/>
-
-    <!-- Dangerous Permissions - must be requested at runtime -->
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
-
-    <!-- hasFragileUserData prompts the user with the option to keep app data upon uninstall -->
     <application
         android:name=".BloomReaderApplication"
-        android:requestLegacyExternalStorage="true"
-        android:preserveLegacyExternalStorage="true"
         android:allowBackup="true"
         android:hasFragileUserData="true"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
+        android:preserveLegacyExternalStorage="true"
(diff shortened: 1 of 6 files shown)


hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/284)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`app/src/main/java/org/sil/bloom/reader/ReaderActivity.java` line 34 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/284#-N2Nxf-B3oo8wwWSSHQW:-N2Nxf-B3oo8wwWSSHQX:bfzhmwy) ([raw file](https://github.com/bloombooks/bloomreader/blob/db6bf631776501fad3e7031e383ce946145213a4/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L34)):*
> ```Java
> 
>     public static boolean haveCurrentWebView(WebView browser) {
>         String agent = browser.getSettings().getUserAgentString();
> ```

Given your concerns in the following comment, why not just look directly at the version of webview available?
https://stackoverflow.com/questions/29145884/get-webview-version-number-for-lollipop

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/284)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`app/src/main/java/org/sil/bloom/reader/ReaderActivity.java` line 34 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/284#-N2Nxf-B3oo8wwWSSHQW:-N2w70BJ8SdqaHJOb9QG:b-2coimc) ([raw file](https://github.com/bloombooks/bloomreader/blob/db6bf631776501fad3e7031e383ce946145213a4/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L34)):*

Sorry, I don't see that your latest commit addressed my question. Instead, it seems to have added further user-agent based matching.
</blockquote></details>

To answer the original question, it was mainly because it was very difficult to find out what version of Chrome we really require! But it's a good thing you asked, because my attempts to test it verified that even systems as late as Android 6 (and having the "wv" in their agent strings) may have WebView versions that don't work. The nearest I could pin down is that Chrome 51 works and Chrome 44 doesn't. So I've set the bar at 51. That might cause some unnecessary upgrade prompts, but it will be a good thing for the user to do anyway.

On the second question, I believe the user agent is a pretty good indication of what version of Chrome/WebView we're getting. That's what the article you referred me to said to do in the first answer...the other answers appear to have variable applicability to different versions of Android.

hatton (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/284)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @hatton and @JohnThomson)

___
*[`app/src/main/java/org/sil/bloom/reader/ReaderActivity.java` line 67 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/284#-N3PP2gHDGaXiY_QkGVq:-N3PP2gHDGaXiY_QkGVr:b-z6o3wb) ([raw file](https://github.com/bloombooks/bloomreader/blob/db6bf631776501fad3e7031e383ce946145213a4/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L67)):*
> ```Java
>             Intent intent = new Intent(this, NeedNewerWebViewActivity.class);
>             startActivity(intent);
>             // We'll continue normally, but if the user chooses 'back' he will be in the
> ```

Can you add a comment saying why we want to continue normally instead of quitting?



_Code quote (from app/src/main/res/values/strings.xml):_
```XML
 <string name="need_newer_webview">In order to use this app, you will need to update the <a href="https://play.google.com/store/apps/details?id=com.google.android.webview">Android System WebView</a> from the Google Play Store.</string>
```

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/284)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @hatton)

___
*[`app/src/main/java/org/sil/bloom/reader/ReaderActivity.java` line 67 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/284#-N3PP2gHDGaXiY_QkGVq:-N3Q0__K91VHmt84r4W3:bwe52c7) ([raw file](https://github.com/bloombooks/bloomreader/blob/db6bf631776501fad3e7031e383ce946145213a4/app/src/main/java/org/sil/bloom/reader/ReaderActivity.java#L67)):*

Can you add a comment saying why we want to continue normally instead of quitting?
</blockquote></details>

I've tried. Not sure it actually IS the right thing to do...I suppose by 'quitting' you mean abandon starting up the reader activity, so going back from the warning will get you to the main screen? I think that could be done...

PR #289, opened 2022-08-05
Don't crash when trying to download an invalid .bloomPUB (BL-11420)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/289)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/DownloadsView.java
+++ b/app/src/main/java/org/sil/bloom/reader/DownloadsView.java
@@ -12,6 +12,7 @@ import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewParent;
 import android.widget.LinearLayout;
+import android.widget.Toast;
 
 import org.sil.bloom.reader.models.BookCollection;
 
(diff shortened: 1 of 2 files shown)


gmartin7 (changes requested):
Just one question... otherwise :lgtm: 

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/289#-:-N8jj8tK3nflFfXYiQ_b:b-j99dl)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`app/src/main/java/org/sil/bloom/reader/ZipFileOrUri.java` line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/289#-N8jih1CEVwknjqfBa1-:-N8jih1CEVwknjqfBa10:bnnm4rm) ([raw file](https://github.com/bloombooks/bloomreader/blob/df064b1ef52fc17eae40cf15e7057599da08d31e/app/src/main/java/org/sil/bloom/reader/ZipFileOrUri.java#L32)):*
> ```Java
>         try {
>             outputDir = output;
>             zipFile = new ZipFile(input);
> ```

Is there a reason for the change in order? Do you need to make sure the outputDir is set before you run into a zip file io exception? Or are you catching an outputDir problem before creating a zipfile?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/289)* status: all files reviewed, 1 unresolved discussion (waiting on @gmartin7)

___
*[`app/src/main/java/org/sil/bloom/reader/ZipFileOrUri.java` line 32 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/289#-N8jih1CEVwknjqfBa1-:-N8jlZE71Vl7Um_Xu7FO:b-n2axui) ([raw file](https://github.com/bloombooks/bloomreader/blob/df064b1ef52fc17eae40cf15e7057599da08d31e/app/src/main/java/org/sil/bloom/reader/ZipFileOrUri.java#L32)):*
<details><summary><i>Previously, gmartin7 (Gordon Martin) wrote…</i></summary><blockquote>

Is there a reason for the change in order? Do you need to make sure the outputDir is set before you run into a zip file io exception? Or are you catching an outputDir problem before creating a zipfile?
</blockquote></details>

JT and I thought it was a slight improvement to set the outputDir before ignoring an exception while trying to create the ZipFile. Probably we should be handling the zipfile error better generally (rather than ignoring it), but this change in order might help in some situation. (It would have helped a little in ours where the invalid zip file was causing a subsequent null pointer exception when trying to use the null outputDir.)

PR #295, opened 2022-08-11
Open Book Detail link in external browser (BL-11409)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/295)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java
+++ b/app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java
@@ -2,6 +2,7 @@ package org.sil.bloom.reader;
 
 import android.app.Activity;
 import android.content.Context;
+import android.content.Intent;
 import android.net.Uri;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
@@ -42,14 +43,20 @@ public class BloomLibraryWebViewClient extends WebViewClient {
         mOwner = owner;
     }
 
-    // Without this, or at least without this class, we get an instance of Chrome loaded.
-    // Review: are there any links that we want to launch a browser? Maybe if URL does not
-    // start with bloomlibrary.org/app-hosted-v1 we should return true? e.g., might we want to open the
-    // Full details view in Chrome? Or some of the links in book details?
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
-        //return super.shouldOverrideUrlLoading(view, request);
-        return false;
+        // Note, the webview never sees links when they are handled by the blorg router.
+
+        final Uri uri = request.getUrl();
+        if (String.valueOf(uri).contains("app-hosted-")) {
+            // load this url in the webView itself
+            return false;
+        } else {
+            // Open in external browser if not an app-hosted link (e.g. normal book detail page)
+            final Intent intent = new Intent(Intent.ACTION_VIEW, uri);
+            mOwner.startActivity(intent);
+            return true;
+        }
     }
 
     class Response {


JohnThomson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/295)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java` line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/295#-N9DRioE21KQkVgP8sMw:-N9DRioE21KQkVgP8sMx:bqy0y18) ([raw file](https://github.com/bloombooks/bloomreader/blob/da9f62ab22b5db8f76c04a59a9d4b9e66a3e4e55/app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java#L57)):*
> ```Java
>             // Open in external browser if not an app-hosted link (e.g. normal book detail page)
>             final Intent intent = new Intent(Intent.ACTION_VIEW, uri);
>             mOwner.startActivity(intent);
> ```

I'm not clear why the two lines above are needed. Doesn't returning true alone cause the link to open in the system browser?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/295)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[`app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java` line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/295#-N9DRioE21KQkVgP8sMw:-N9DzMsL3V3rv3vrmho8:biklhac) ([raw file](https://github.com/bloombooks/bloomreader/blob/da9f62ab22b5db8f76c04a59a9d4b9e66a3e4e55/app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java#L57)):*

I'm not clear why the two lines above are needed. Doesn't returning true alone cause the link to open in the system browser?
</blockquote></details>

No, doc says, "returning true causes the current WebView to abort loading the URL, while returning false causes the WebView to continue loading the URL as usual."

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/295)* status: all files reviewed, 1 unresolved discussion

___
*[`app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java` line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/295#-N9DRioE21KQkVgP8sMw:-N9Yfrm-6JW1cRr1A8Yg:b-72vnob) ([raw file](https://github.com/bloombooks/bloomreader/blob/da9f62ab22b5db8f76c04a59a9d4b9e66a3e4e55/app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java#L57)):*

No, doc says, "returning true causes the current WebView to abort loading the URL, while returning false causes the WebView to continue loading the URL as usual."
</blockquote></details>

Yes, but what happens next when the current WebView aborts loading the URL? When I didn't have this class at all, it used the system browser to open it instead, and I was guessing that's what happens after it aborts loading it in this one. Of course, not installing a WebViewClient at all may be different from either returning true or returning false.

Anyway, what you have apparently works, so I guess all is well.

andrew-polk:
*[`app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java` line 57 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/295#-N9DRioE21KQkVgP8sMw:-N9YhXca1ZeQ0l3oHzJ6:br94wt9) ([raw file](https://github.com/bloombooks/bloomreader/blob/da9f62ab22b5db8f76c04a59a9d4b9e66a3e4e55/app/src/main/java/org/sil/bloom/reader/BloomLibraryWebViewClient.java#L57)):*

Yes, but what happens next when the current WebView aborts loading the URL? When I didn't have this class at all, it used the system browser to open it instead, and I was guessing that's what happens after it aborts loading it in this one. Of course, not installing a WebViewClient at all may be different from either returning true or returning false.

Anyway, what you have apparently works, so I guess all is well.
</blockquote></details>

I'm pretty sure I had tried it with just returning true and it didn't do anything.

PR #303, opened 2022-10-05
Make analytics events for downloading a book (BL-11423)

JohnThomson:
This change is [Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/303)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/DownloadsView.java
+++ b/app/src/main/java/org/sil/bloom/reader/DownloadsView.java
@@ -14,6 +14,9 @@ import android.view.ViewParent;
 import android.widget.LinearLayout;
 import android.widget.Toast;
 
+import com.segment.analytics.Analytics;
+import com.segment.analytics.Properties;
+
 import org.sil.bloom.reader.models.BookCollection;
 
 import java.io.File;
@@ -293,6 +296,7 @@ public class DownloadsView extends LinearLayout {
                 File downloadDest = new File(getDownloadDir(), fileName + ".bloompub");
                 long downloadId = cursor.getLong(cursor.getColumnIndex(DownloadManager.COLUMN_ID));
                 int downloadStatus = cursor.getInt(cursor.getColumnIndex(DownloadManager.COLUMN_STATUS));
+                String downloadDescription = cursor.getString(cursor.getColumnIndex(DownloadManager.COLUMN_DESCRIPTION));
                 switch (downloadStatus) {
                     default:
                         // if we see a failed download, for now we'll ignore it.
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/sil/bloom/reader/DownloadsView.java
```diff
@@ -446,6 +449,12 @@ private void handleDownloadComplete(String downloadDestPath, long downloadId) {
         String fileName = DownloadsView.getFileNameFromUri(downloadDestPath);
         File dest = new File(BookCollection.getLocalBooksDirectory(), fileName + ".bloompub");
         IOUtilities.copyFile(source.getPath(), dest.getPath());
+        Properties props = new Properties();
+        BloomFileReader reader = new BloomFileReader(mContext, dest.getPath());
+        props.putValue("bookInstanceId", reader.getStringMetaProperty("bookInstanceId", ""));
+        props.putValue("title", reader.getStringMetaProperty("title", ""));
+        props.putValue("originalTitle", reader.getStringMetaProperty("originalTitle", ""));
```
andrew-polk, line 456: Do we have easy access to the parse ID? I think we will want that to easily tie to the notifications later.
Plus, while there is a one-to-one correspondence with bookInstanceId in the ideal world, there isn't necessarily in the real world.

I also wondered if we will want anything about the language. But maybe that's too complicated...
In theory, knowing the book, we know the languages in the book, but that doesn't tell us which language the user was searching within.
I guess if we have easy access to the language they are searching in, we should add it, but my guess is we don't have that.

andrew-polk (changes requested):

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/303)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`app/src/main/java/org/sil/bloom/reader/DownloadsView.java` line 456 at r1](https://reviewable.io/reviews/bloombooks/bloomreader/303#-NDd6d1IdGaXvrrmOxkC-r1-456:-NDdRuX6506HAPbyHZkS:b-896fix) ([raw file](https://github.com/bloombooks/bloomreader/blob/36eddc7f5f41fc9d1fc58a48da7a9efec097ad64/app/src/main/java/org/sil/bloom/reader/DownloadsView.java#L456)):*

Do we have easy access to the parse ID? I think we will want that to easily tie to the notifications later.
Plus, while there is a one-to-one correspondence with bookInstanceId in the ideal world, there isn't necessarily in the real world.

I also wondered if we will want anything about the language. But maybe that's too complicated...
In theory, knowing the book, we know the languages in the book, but that doesn't tell us which language the user was searching within.
I guess if we have easy access to the language they are searching in, we should add it, but my guess is we don't have that.
</blockquote></details>

Done.

andrew-polk (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/bloombooks/bloomreader/303)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

PR #313, opened 2025-02-20
Use ICU sort for book titles in book list (BL-14289)

andrew-polk:
Note that Android < 7.0 will still use the old sort since ICU was introduced in Android in 7.0 (api 24).

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/313)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java
+++ b/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java
@@ -32,6 +32,9 @@ package org.sil.bloom.reader.models;
  */
 
 import java.util.Comparator;
+import java.util.Locale;
+import android.icu.text.Collator;
+import android.os.Build;
 
 /**
  * This is an updated version with enhancements made by Daniel Migowski,
@@ -44,6 +47,16 @@ import java.util.Comparator;
  *   Collections.sort(your list, new AlphanumComparator());
  */
 public class AlphanumComparator implements Comparator<BookOrShelf> {
+
+    private static Collator sIcuCollator = null;
+
+    static {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+            sIcuCollator = Collator.getInstance(Locale.ROOT);
+            sIcuCollator.setStrength(Collator.SECONDARY);
+        }
+    }
+
     private final boolean isDigit(char ch) {
         return ((ch >= 48) && (ch <= 57)); // Digits 0 through 9
     }
@@ -177,7 +190,11 @@ public class AlphanumComparator implements Comparator<BookOrShelf> {
                     }
                 }
             } else {
-                result = thisChunk.compareToIgnoreCase(thatChunk);
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
+                    result = sIcuCollator.compare(thisChunk, thatChunk);
+                } else {
+                    result = thisChunk.compareToIgnoreCase(thatChunk);
+                }
             }
 
             if (result != 0)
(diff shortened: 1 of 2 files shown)


JohnThomson (approved):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/313)* status: 1 of 2 files reviewed, 2 unresolved discussions

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomReader/313#-OJ_depaDVEHU5EaDWH5:-OJ_depaDVEHU5EaDWH6:b-cwsycd) (no related file):*
All looks good. I had a couple of suggestions to consider if you want to have another go at the test, but it's simple code and we have plenty that we haven't been able to figure unit tests for. So I'm approving and leaving it up to you whether to try again or just merge.
___
*[`app/src/test/java/org/sil/bloom/reader/models/BookOrShelfTest.java` line 78 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/313#-OJ_bpzE00eeTd2cUFlb:-OJ_bpzE00eeTd2cUFlc:b1k974k) ([raw file](https://github.com/BloomBooks/BloomReader/blob/1aa0c0c1b1c64fd7a93af3f56ec75159c2adbf7b/app/src/test/java/org/sil/bloom/reader/models/BookOrShelfTest.java#L78)):*
> ```Java
>     // At this point, I'm giving up and saying it isn't worth it.
>     // The complication is the gated code in AlphanumComparator for the SDK version.
>     // When running unit tests, Build.Version.SDK_INT is 0.
> ```

What if you change the test in the actual compare function to if (sIcuCollator != null)? Then you could wrap the code that initializes it in a new method (without any condition) and call that from the test. (Presumably the test environment does actually support making one?)
Another option would be to make a static IsIcuCollationSupported and test that everywhere. Then you just need a 'real' line that initializes it in production and something in the test that just sets it. (In fact you could test both ways if you want, as the method now does.)

JohnThomson (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/313)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/313)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[`app/src/test/java/org/sil/bloom/reader/models/BookOrShelfTest.java` line 78 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/313#-OJ_bpzE00eeTd2cUFlb:-OJ_rrt6BJ2UYuK9O6GD:b-61erwd) ([raw file](https://github.com/BloomBooks/BloomReader/blob/1aa0c0c1b1c64fd7a93af3f56ec75159c2adbf7b/app/src/test/java/org/sil/bloom/reader/models/BookOrShelfTest.java#L78)):*

What if you change the test in the actual compare function to if (sIcuCollator != null)? Then you could wrap the code that initializes it in a new method (without any condition) and call that from the test. (Presumably the test environment does actually support making one?)
Another option would be to make a static IsIcuCollationSupported and test that everywhere. Then you just need a 'real' line that initializes it in production and something in the test that just sets it. (In fact you could test both ways if you want, as the method now does.)
</blockquote></details>

Thanks. Yes, those could probably help us get past the first complication.
I've added more comments before the commented out test.
Let me know if that's makes sense.

(Another piece of this which probably isn't worth commenting on in code because it can easily be ignored with annotations is...
lint will only accept certain ways of the production code handling the >= 24 gate. 
So your suggestions probably wouldn't pass lint. : (

PR #316, opened 2025-02-21
Use title to sort books rather than file name (BL-14289)

andrew-polk:
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/316)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/BookListAdapter.java
+++ b/app/src/main/java/org/sil/bloom/reader/BookListAdapter.java
@@ -74,8 +74,7 @@ public class BookListAdapter extends RecyclerView.Adapter<BookListAdapter.ViewHo
             holder.linearLayout.setBackgroundColor(ContextCompat.getColor(holder.getContext(), R.color.new_book_highlight));
         else
             holder.linearLayout.setBackgroundColor(Color.WHITE);
-        if ("loadExternalFiles".equals(holder.bookOrShelf.specialBehavior) || "importOldBloomFolder".equals(holder.bookOrShelf.specialBehavior)
-                || "getBooksFromLibrary".equals(holder.bookOrShelf.specialBehavior)) {
+        if ("loadExternalFiles".equals(holder.bookOrShelf.specialBehavior) || "importOldBloomFolder".equals(holder.bookOrShelf.specialBehavior)) {
             // I'm not sure this is the best way to do this. Maybe we should use an entirely different layout?
             // I'm not getting the exact appearance JohnH put in the doc...the image he gave me seems to have
             // only two 'contacts' instead of three, and it's bigger than the mockup. It may also be higher than
--- a/app/src/main/java/org/sil/bloom/reader/SetBookListItemViewExtrasTask.java
+++ b/app/src/main/java/org/sil/bloom/reader/SetBookListItemViewExtrasTask.java
@@ -61,18 +61,6 @@ public class SetBookListItemViewExtrasTask extends AsyncTask<Void, Void, SetBook
             viewHolder.imageView.setPadding(newPadding, newPadding, newPadding, newPadding);
             return;
         }
-        if ("getBooksFromLibrary".equals(bookOrShelf.specialBehavior)) {
-            viewHolder.imageView.setImageResource(R.drawable.ic_search);
-            final Resources resources = viewHolder.getContext().getResources();
-            // Converts 'dp' measurements into actual pixels
-            int newPadding = (int) TypedValue.applyDimension(
-                    TypedValue.COMPLEX_UNIT_DIP,
-                    12,
-                    resources.getDisplayMetrics()
-            );
-            viewHolder.imageView.setPadding(newPadding, newPadding, newPadding, newPadding);
-            return;
-        }
         if ("importOldBloomFolder".equals(bookOrShelf.specialBehavior)) {
             // Enhance: add some suitable icon
             //viewHolder.imageView.setImageResource(R.drawable.ic_sd_card);
--- a/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java
+++ b/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java
@@ -40,7 +40,7 @@ import android.os.Build;
  * This is an updated version with enhancements made by Daniel Migowski,
  * Andre Bogus, and David Koelle. Updated by David Koelle in 2017.
  *
- * Some changes were made by the Bloom team for their needs in 2020.
+ * Some changes were made by the Bloom team for their needs.
  *
  * To use this class:
  *   Use the static "sort" method from the java.util.Collections class:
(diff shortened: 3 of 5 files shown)


andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/316)* status: 0 of 3 files reviewed, all discussions resolved

___
*[`app/src/main/java/org/sil/bloom/reader/BookListAdapter.java` line 78 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJdpfE409EjOpy38-e9:-OJdpfE409EjOpy38-eA:bxxri55) ([raw file](https://github.com/BloomBooks/BloomReader/blob/c0d513d22b5d7dab13284b759bf585098b55ce83/app/src/main/java/org/sil/bloom/reader/BookListAdapter.java#L78)):*
> ```Java
>             holder.linearLayout.setBackgroundColor(Color.WHITE);
>         if ("loadExternalFiles".equals(holder.bookOrShelf.specialBehavior) || "importOldBloomFolder".equals(holder.bookOrShelf.specialBehavior)
>                 || "getBooksFromLibrary".equals(holder.bookOrShelf.specialBehavior)) {
> ```

This change is unrelated to the fix.
I realized we don't use this special case. The books from blorg UI element was actually implemented outside of the list (so it doesn't scroll, for one thing).
So this is just unrelated cleanup.
___
*[`app/src/main/java/org/sil/bloom/reader/SetBookListItemViewExtrasTask.java` line 75 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJdpsnI2pvb5X65EG7M:-OJdpsnI2pvb5X65EG7N:bxxri55) ([raw file](https://github.com/BloomBooks/BloomReader/blob/c0d513d22b5d7dab13284b759bf585098b55ce83/app/src/main/java/org/sil/bloom/reader/SetBookListItemViewExtrasTask.java#L75)):*
> ```Java
>             viewHolder.imageView.setPadding(newPadding, newPadding, newPadding, newPadding);
>             return;
>         }
> ```

This change is unrelated to the fix.
I realized we don't use this special case. The books from blorg UI element was actually implemented outside of the list (so it doesn't scroll, for one thing).
So this is just unrelated cleanup.

JohnThomson (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/316)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[a discussion](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJeK0zr7cheaeugB74t:-OJeK0zr7cheaeugB74u:b-z8652v) (no related file):*
Up to you how much is worth fixing. The main problems I spotted are pre-existing.
___
*[`app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java` line 126 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJeIEOE0-V7xsZ-6CPZ:-OJeIEOE0-V7xsZ-6CP_:b-xxoazx) ([raw file](https://github.com/BloomBooks/BloomReader/blob/e85795ad3b49627e86265380a85bb8fa55798c73/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java#L126)):*
> ```Java
>             // unless it has sortLast set (not used as of Feb 2025).
>             // Note: if both are special, sort by name.
>             if (one.specialBehavior == null) {
> ```

This comment seems out of place. The following lines are about sortLast.
More seriously, this will have problems if both are sortLast. I think it is a requirement that a compare function must be symmetric. But here, if two elements are both special and both sortLast, whichever one happens to be "one" will sort first.
(Of course it doesn't matter much if we're not using sortLast. But let's not leave something incorrectly implemented. It should be either fixed, or the sortLast special case removed or commented out.)
(And I now see this isn't a new problem. And perhaps the idea all along was to only support one sortLast element, which would be OK.)
___
*[`app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java` line 137 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJeJhyd5-He_9IhTZR9:-OJeJhyeDCYXXQS0WjLz:b-xzp8us) ([raw file](https://github.com/BloomBooks/BloomReader/blob/e85795ad3b49627e86265380a85bb8fa55798c73/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java#L137)):*
> ```Java
>         if (s1 == null ^ s2 == null) {
>             return (s1 == null) ? 1 : -1;
>         }
> ```

Again, not new, but this has the same probably-never-happen problem as above: if both are null, the result is not symmetric. Better to answer zero for that case?

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/316)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on @JohnThomson)

___
*[`app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java` line 126 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJeIEOE0-V7xsZ-6CPZ:-OJecizAELoVtGq7cc4r:b-22kqd1) ([raw file](https://github.com/BloomBooks/BloomReader/blob/e85795ad3b49627e86265380a85bb8fa55798c73/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java#L126)):*

This comment seems out of place. The following lines are about sortLast.
More seriously, this will have problems if both are sortLast. I think it is a requirement that a compare function must be symmetric. But here, if two elements are both special and both sortLast, whichever one happens to be "one" will sort first.
(Of course it doesn't matter much if we're not using sortLast. But let's not leave something incorrectly implemented. It should be either fixed, or the sortLast special case removed or commented out.)
(And I now see this isn't a new problem. And perhaps the idea all along was to only support one sortLast element, which would be OK.)
</blockquote></details>

I decided to just get rid of sortLast since we aren't using it.
___
*[`app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java` line 137 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJeJhyd5-He_9IhTZR9:-OJe_YK0CA9dCNxhoFxs:bac5b8j) ([raw file](https://github.com/BloomBooks/BloomReader/blob/e85795ad3b49627e86265380a85bb8fa55798c73/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java#L137)):*

Again, not new, but this has the same probably-never-happen problem as above: if both are null, the result is not symmetric. Better to answer zero for that case?
</blockquote></details>

If both are null, it falls through to sorting on pathOrUri. I think that is expected/desired.

JohnThomson (approved):
:lgtm:

> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-:-OJyvDtiD4lfBXKdkAd3:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @andrew-polk)

___
*[`app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java` line 137 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/316#-OJeJhyd5-He_9IhTZR9:-OJyuvuEBUZAOj9SfYPG:b-4s6ja0) ([raw file](https://github.com/BloomBooks/BloomReader/blob/e85795ad3b49627e86265380a85bb8fa55798c73/app/src/main/java/org/sil/bloom/reader/models/AlphanumComparator.java#L137)):*

If both are null, it falls through to sorting on pathOrUri. I think that is expected/desired.
</blockquote></details>

Sorry, you're right, it's an exclusive or.

PR #320, opened 2025-06-19
Added support for sharing multiple books/shelves at once (BL-14392)

NJStriebel:
Unfortunately, I haven't yet found a fix for the bug that specifically prevents sharing "The Moon and the Cap."

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/320)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/BookListAdapter.java
+++ b/app/src/main/java/org/sil/bloom/reader/BookListAdapter.java
@@ -9,6 +9,7 @@ import androidx.recyclerview.widget.RecyclerView;
 import android.text.TextUtils;
 import android.util.TypedValue;
 import android.view.LayoutInflater;
+import android.view.MenuItem;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/320)* status: all files reviewed, 1 unresolved discussion (waiting on @NJStriebel)

___
*[`app/src/main/java/org/sil/bloom/reader/MainActivity.java` line 928 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/320#-OT9E6a5E6taa7cQOs2t:-OTDBHzcElDTo6HSF1ST:b-dnx1li) ([raw file](https://github.com/BloomBooks/BloomReader/blob/de1a91fc5002de5c439168549517797a95873e9a/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L928)):*
> When I was writing the code, I thought about your other question, and to the best of my understanding, calling "notifyDataSetChanged" at that point would be redundant. Calling "closeContextualActionBar" just a couple lines later results in a call to "contextualActionBarMode.finish()" which implicitly calls "onDestroyActionMode" calling "mBookListAdapter.clearSelection()" which would have another "notifyDataSetChanged" method call in response to the same change.

> I think it's also redundant in the deleteShelf and deleteBook methods, but I was hesitant to make more modifications than I needed to in any code that was there before I arrived. Do you want me to remove those redundant calls?

As we discussed...
I definitely understand making minimal changes to code you don't understand. 
However given that
1. You've already done the analysis someone else would later have to do to clean it up.
2. This code is about to be tested with the new feature.

Let's do the clean up now.

Thanks!

PR #321, opened 2025-06-20
Add current Bloom Reader version to the top of the about popup

NJStriebel:
…14406)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/321)

### DIFF
--- a/app/src/main/java/org/sil/bloom/reader/MainActivity.java
+++ b/app/src/main/java/org/sil/bloom/reader/MainActivity.java
@@ -550,16 +550,11 @@ public class MainActivity extends BaseActivity
         // Insert the build version and date into the appropriate control.
         // We have to find it indirectly through the navView's header or it won't be found
         // this early in the view construction.
-        TextView versionDate = navigationView.getHeaderView(0).findViewById(R.id.versionDate);
-        try {
-            String versionName = getPackageManager().getPackageInfo(getPackageName(), 0).versionName;
-            Date buildDate = new Date(BuildConfig.TIMESTAMP);
-            String date = DateFormat.format("dd MMM yyyy", buildDate).toString();
-            versionDate.setText(getVersionAndDateText(versionName, date));
-        } catch (PackageManager.NameNotFoundException e) {
-            e.printStackTrace();
+        String versionDateText = getVersionAndDateText();
+        if(versionDateText != null) {
+            TextView versionDateUIElement = navigationView.getHeaderView(0).findViewById(R.id.versionDate);
+            versionDateUIElement.setText(versionDateText);
         }
-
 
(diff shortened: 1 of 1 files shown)


andrew-polk (changes requested):
Looks good overall. I suggested a couple tweaks.

> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/321#-:-OTEHoLh7a4yh-iMkUz-:b-nynj09)* status: all files reviewed, 2 unresolved discussions (waiting on @NJStriebel)

___
*[`app/src/main/java/org/sil/bloom/reader/MainActivity.java` line 637 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/321#-OTEH-R44jG_jRqsa8vI:-OTEH-R44jG_jRqsa8vJ:b-wa96i3) ([raw file](https://github.com/BloomBooks/BloomReader/blob/7e2dad88b3c19a80f8be39603d48950a978b7250/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L637)):*
> ```Java
>             e.printStackTrace();
>         }
> ```

In the current code, if there is an exception, we just don't call `versionDate.setText()`. 
Now, we will set it to ", ".
I think we should have `getVersionAndDateText()` return null if there is a problem. Then the caller can decide what to do. In createMainActivity, it won't call `versionDate.setText()`. In `versionDate.setText()`, ideally it wouldn't add the `\n`.
___
*[`app/src/main/java/org/sil/bloom/reader/MainActivity.java` line 1148 at r1](https://reviewable.io/reviews/BloomBooks/BloomReader/321#-OTEGQtx3pPWka20isgf:-OTEGQtx3pPWka20isgg:bmzotmj) ([raw file](https://github.com/BloomBooks/BloomReader/blob/7e2dad88b3c19a80f8be39603d48950a978b7250/app/src/main/java/org/sil/bloom/reader/MainActivity.java#L1148)):*
> ```Java
>                 showLocationAnalyticsData();
>         } else if (id == R.id.about_reader) {
>                 String aboutTitle = "About Bloom Reader\n" + getVersionAndDateText();
> ```

You may have gotten confused by the IDE which attempts to help you by masking what is happening, but `getString(R.string.about_bloom_reader)` got replaced by the raw string `About Bloom Reader`. We need to maintain the getString lookup for localization.

andrew-polk (approved):
What you did looks good.
It is unfortunate that Java (if I remember) doesn't allow for optional parameters.

> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/321#-:-OTYF34J7KS717t76U4d:b-7d2ksr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @NJStriebel)

PR #330, opened 2026-07-16
Auto-resolve alpha version collisions on commit-less rebuilds

andrew-polk:
## Problem

Alpha is sometimes rebuilt with no new commit — most commonly when a bloom-player update dispatches the publish workflow. The git-derived patch number (commits since the last version bump) then repeats, so the publish fails with a duplicate-versionCode rejection from Play. This just happened in practice.

## Fix (final shape)

**Upgrade gradle-play-publisher 2.8.0 → 3.13.0** and let the plugin handle collisions natively: the alpha flavor sets `resolutionStrategy = AUTO` (scoped via `playConfigs`), which publishes **max(derived versionCode, highest on Play + 1)**. The version name is kept in sync with the resolved code via the plugin's documented `androidComponents.onVariants` pattern, and the workflow reads the actually-published version from `output-metadata.json` for the artifact name, tag, and bloomlibrary.org version file.

- **Production deliberately keeps the strict default (FAIL)**: it is never rebuilt without a commit, so a duplicate versionCode there means something is wrong and should fail loudly.
- AUTO is **gated on Play credentials being present**, so local builds don't acquire a Play API dependency.
- The AUTO floor makes the +77 legacy TeamCity offset redundant (removed) and neutralizes the "editing the version lines resets the counter" hazard for alpha.
- Migration notes: `publishAlphaRelease` → `publishAlphaReleaseApps`; `serviceAccountCredentials` is now optional (dummy.json fallback removed); retain and the retain-guard use the Property API.

## How it got here (for reviewers reading the commit history)

1. `13505ef` tried GPP **2.8's** `resolutionStrategy = "auto"` — review caught, and a local probe confirmed, that 2.8's execution-time overrides throw under AGP 8 ("The value for this property cannot be changed any further"); every CI publish would have failed.
2. `c5400d8`/`c16ad01` replaced it with a hand-rolled Play API query (JWT + `edits/apks.list`) in the workflow — correct but ~60 lines of bash.
3. `9ebaccb` upgraded the plugin to 3.13.0, whose AUTO works properly on AGP 8, and deleted the bash. Net simpler than where we started.

## Verification

- Locally: gradle configures on 3.13; the alpha versionName wiring verified end-to-end on a real build (`3.4.42-alpha` / `304042` in output-metadata.json); the retain guard reads `[104001]` off the real publish task; the credentials gate yields FAIL locally; task graphs for both flavors resolve; full unit-test suite green.
- End-to-end proof still requires a real run: after merging, re-dispatch the workflow with no new commit (the exact scenario that failed) and confirm it publishes highest+1 with a matching name.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

- - -
This change is [Reviewable](https://reviewable.io/reviews/BloomBooks/BloomReader/330)

### DIFF
--- a/.github/workflows/play-publish.yml
+++ b/.github/workflows/play-publish.yml
@@ -27,11 +27,12 @@
 # Versioning: the patch number (3.4.NNN) is the number of commits since
 # versionMajor/versionMinor last changed in app/build.gradle, so it resets to 0
 # automatically when the version is bumped, and master (alpha) and release
-# (production) count independently. Alpha additionally floors its number at
-# (Play's highest uploaded versionCode + 1), so rebuilds with no new commit
-# (e.g. dispatched after a bloom-player update) still publish. NOTE:
-# build.gradle computes versionCode = major*100000 + minor*1000 + patch, so
-# patch must stay below 1000.
+# (production) count independently. For alpha, gradle-play-publisher's
+# resolutionStrategy AUTO (see playConfigs in app/build.gradle) additionally
+# floors the published versionCode at (Play's highest uploaded code + 1), so
+# rebuilds with no new commit (e.g. dispatched after a bloom-player update)
+# still publish. NOTE: build.gradle computes versionCode = major*100000 +
+# minor*1000 + patch, so patch must stay below 1000.
 
 name: Publish to Play Console
 
@@ -112,7 +113,6 @@ jobs:
               exit 1
             fi
             BUILD_NUMBER="$PATCH_INPUT"
-            echo "patch_overridden=true" >> "$GITHUB_OUTPUT"
           else
             BUMP_COMMIT=$(git log -1 --format=%H -G'^def version(Major|Minor)' -- app/build.gradle)
             if [ -z "$BUMP_COMMIT" ]; then
@@ -136,7 +136,7 @@ jobs:
             # The full "build" deliberately compiles, lints, and unit-tests
             # every flavor and build type — each alpha run verifies that
             # nothing is broken even in the variants we don't publish.
-            TASKS="build publishAlphaRelease promoteAlphaReleaseArtifact"
+            TASKS="build publishAlphaReleaseApps promoteAlphaReleaseArtifact"
           else
             # APK-only for production: the "Bloom Reader" store listing is
             # managed in the Play Console, and the listing text/graphics in
@@ -148,8 +148,6 @@ jobs:
           MINOR=$(sed -n 's/^def versionMinor = \([0-9]*\).*/\1/p' app/build.gradle)
           echo "flavor=$FLAVOR" >> "$GITHUB_OUTPUT"
           echo "flavor_lc=${FLAVOR,,}" >> "$GITHUB_OUTPUT"
-          echo "major=$MAJOR" >> "$GITHUB_OUTPUT"
-          echo "minor=$MINOR" >> "$GITHUB_OUTPUT"
           echo "build_number=$BUILD_NUMBER" >> "$GITHUB_OUTPUT"
           echo "tasks=$TASKS" >> "$GITHUB_OUTPUT"
           echo "Derived version $MAJOR.$MINOR.$BUILD_NUMBER for flavor $FLAVOR (tasks: $TASKS)"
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -120,6 +120,27 @@ android {
...
+            // Gated on credentials because "auto" wires a Play API call into
+            // the alpha release build, which would break local builds that
+            // have no credentials (and no ability to publish anyway).
+            resolutionStrategy = keystoreProperties ? "auto" : "fail"
+            // Runs after resolution: keep the visible version name in sync
+            // with the (possibly bumped) final version code.
+            outputProcessor { output ->
+                int code = output.versionCodeOverride > 0 ? output.versionCodeOverride : output.versionCode
+                output.versionNameOverride = "${code.intdiv(100000)}.${(code % 100000).intdiv(1000)}.${code % 1000}-alpha"
+            }
```
andrew-polk: [Devin] **Investigate**: Auto-bump relies on unverified GPP ordering: outputProcessor must run after version resolution

The outputProcessor (app/build.gradle:139-142) computes the visible version name from `output.versionCodeOverride > 0 ? output.versionCodeOverride : output.versionCode`. This is only correct if gradle-play-publisher 2.8 applies resolutionStrategy = "auto" (which sets versionCodeOverride to the bumped code) BEFORE invoking the outputProcessor. If the processor runs first (or before Play resolution has set the override), versionCodeOverride would still be 0 and the recomputed name would reflect the un-bumped derived code, silently mismatching the code actually published. The PR description explicitly states the live auto-bump path can only be proven by a real run, so this ordering assumption is unverified. Worth confirming against GPP 2.8 source (the comment claims the processor "runs after resolution").

.github/workflows/play-publish.yml
```diff
@@ -209,6 +209,18 @@ jobs:
...
+      # app/build.gradle), so a rebuild with no new commits publishes a
+      # higher version than the git-derived one. Read the final values from
+      # the build output.
+      - name: Read published version
+        id: published
+        run: |
+          META="app/build/outputs/apk/${{ steps.config.outputs.flavor_lc }}/release/output-metadata.json"
+          VERSION_NAME=$(jq -r '.elements[0].versionName' "$META")
+          echo "version_name=$VERSION_NAME" >> "$GITHUB_OUTPUT"
+          echo "Published versionName $VERSION_NAME (versionCode $(jq -r '.elements[0].versionCode' "$META"))"
```
andrew-polk, line 218: [Devin] **Investigate**: Reading versionName from output-metadata.json assumes GPP writes the resolved code back to AGP metadata

The new Read published version step (play-publish.yml:216-222) reads versionName/versionCode from AGP's output-metadata.json and uses the versionName for the uploaded artifact name (play-publish.yml:266). This assumes the bumped versionCode/Name produced by resolutionStrategy = "auto" are reflected in output-metadata.json. If GPP resolves and applies the bumped code only at publish time (not by mutating the assembled APK's metadata that AGP writes during assemble), the artifact name and the log line would show the derived (colliding) version rather than the actually-published one, defeating the stated purpose. The APK baked versionCode vs. what is published to Play is the key thing to confirm on a real run.

app/build.gradle
andrew-polk: [Claude Fable 5] This flag pointed at a real problem, though not quite the one described. The ordering within GPP 2.8 is actually fine (ProcessArtifactMetadata sets the override immediately before invoking the outputProcessor in the same loop). What's broken is deeper: under AGP 8 the variant version properties are locked after configuration, so GPP 2.8's execution-time `versionCodeOverride`/`versionNameOverride` throw 'The value for this property cannot be changed any further' — reproduced locally with a probe task. Every CI alpha publish would have failed. Fixed in c5400d8 by dropping `resolutionStrategy = "auto"` entirely: the workflow now resolves the final patch number against the Play API before gradle runs, so the version is set once at configuration time and no overrides are involved. Resolving.

.github/workflows/play-publish.yml
andrew-polk: [Claude Fable 5] Good instinct — investigating this class of assumption uncovered that the whole mechanism was unsound: under AGP 8, GPP 2.8 cannot apply its execution-time version override at all (the variant version properties are locked after configuration; reproduced locally), so the metadata question was moot — the publish itself would have failed. Fixed in c5400d8: the workflow now resolves the final patch number against the Play API before gradle runs and passes it as -Pbuild.number, so the versionCode/versionName are baked in at configuration time, output-metadata.json trivially matches, and the 'Read published version' step this flag pointed at was removed as unnecessary. Resolving.

andrew-polk:
@greptileai review

andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-07-16 up to commit c5400d8 (two rounds: 13505ef and c5400d8).

Round 1 raised two Investigate flags; investigating them led to the discovery that gradle-play-publisher's execution-time version override cannot work under AGP 8 at all, which c5400d8 fixes by resolving the version against the Play API in the workflow instead. Both threads are replied-to and resolved: [ordering flag](https://github.com/BloomBooks/BloomReader/pull/330#discussion_r3597932909), [metadata flag](https://github.com/BloomBooks/BloomReader/pull/330#discussion_r3597933558).

Round 2 (c5400d8): re-review clean — no bugs, no Investigate flags; Devin marks both prior flags resolved. Its remaining notes are informational (alpha publishing now requires Play API reachability during the version-resolution step, and the max-versionCode query reads apks.list, which would need bundles.list too if we ever switch to AABs).

.github/workflows/play-publish.yml
```diff
@@ -200,13 +198,68 @@ jobs:
...
+            TOKEN=$(curl -sf -X POST https://oauth2.googleapis.com/token \
+              -d "grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&assertion=$HDR.$CLM.$SIG" | jq -r .access_token)
+            if [ -z "$TOKEN" ] || [ "$TOKEN" = "null" ]; then
+              echo "::error::Could not get a Play API token to check for versionCode collisions."
+              exit 1
+            fi
+            BASE="https://androidpublisher.googleapis.com/androidpublisher/v3/applications/org.sil.bloom.reader.alpha"
+            EDIT=$(curl -sf -X POST -H "Authorization: Bearer $TOKEN" "$BASE/edits" | jq -r .id)
+            MAX_CODE=$(curl -sf -H "Authorization: Bearer $TOKEN" "$BASE/edits/$EDIT/apks" | jq '[.apks[]?.versionCode] | max // 0')
+            curl -sf -X DELETE -H "Authorization: Bearer $TOKEN" "$BASE/edits/$EDIT" >/dev/null || true
```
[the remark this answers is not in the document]
andrew-polk: [Claude Fable 5] Fixed in c16ad01: the edit id is now validated and the DELETE runs via an EXIT trap, so the edit is discarded even if the apks-list call fails. One correction for the record: orphaned edits don't actually block later publishes — Play edits are independent and expire on their own, and gradle-play-publisher opens its own edit per run regardless — so the impact was clutter rather than a blocked pipeline. The cleanup is still clearly right. Thanks.

andrew-polk:
@greptileai review

andrew-polk:
@greptileai review

.github/workflows/play-publish.yml
```diff
@@ -209,6 +204,19 @@ jobs:
...
+      # publishes a higher version than the git-derived one. Read the final
+      # values from the build output; for production (strategy FAIL) they
+      # always equal the derived ones.
+      - name: Read published version
+        id: published
+        run: |
+          META="app/build/outputs/apk/${{ steps.config.outputs.flavor_lc }}/release/output-metadata.json"
+          VERSION_NAME=$(jq -r '.elements[0].versionName' "$META")
+          echo "version_name=$VERSION_NAME" >> "$GITHUB_OUTPUT"
+          echo "Published versionName $VERSION_NAME (versionCode $(jq -r '.elements[0].versionCode' "$META"))"
```
andrew-polk, line 218: [Devin] **Investigate**: Alpha artifact name depends on output-metadata.json reflecting the AUTO-resolved code

For alpha, the artifact name and the Read published version step read versionName from output-metadata.json, assuming it contains the AUTO-resolved versionCode rather than the git-derived one. This holds only if gradle-play-publisher wires the resolved code into output.versionCode at assemble time (so the metadata written during build reflects the bumped value). The PR description explicitly notes the end-to-end collision scenario (rebuild with no new commit) was NOT verified — the verified build had derived == resolved (3.4.42-alpha/304042). If the metadata is written with the derived code before the publish task queries Play, the alpha artifact name would show the wrong (derived) version. This affects only the cosmetic artifact name for alpha (production uses FAIL, so derived == published), so it is low-impact, but worth confirming on the first real commit-less rebuild.

README.md
andrew-polk, line 1: [Devin] **Investigate**: README plugin task list still uses gradle-play-publisher 2.x task names

The README task reference still lists publish{Alpha/Beta/Production}Release, ...ReleaseApk, and ...ReleaseListing — the gradle-play-publisher 2.x naming. This PR upgrades to 3.13.0 and the workflow itself switched alpha from publishAlphaRelease to publishAlphaReleaseApps, indicating task names changed in 3.x. These README lines were not updated and may now be stale/misleading for anyone invoking the plugin locally.

.github/workflows/play-publish.yml
andrew-polk: [Claude Fable 5] The mechanism is sound in GPP 3.x: unlike 2.8 (which patched versions at publish-task execution time), 3.13 wires the resolved code into the variant via AGP's new variant API providers at configuration time — the same provider chain feeds the APK manifest, output-metadata.json, and the upload. The local verification (3.4.42-alpha / 304042) exercised exactly that provider chain end to end; what remains unproven is only the Play-query value feeding it, and Devin is right that the first real commit-less rebuild is the confirmation point — that check is already the PR's documented post-merge verification step, and as noted the worst case is a mislabeled artifact name, not a wrong publish. Resolving; the post-merge dispatch will settle it empirically.

README.md
andrew-polk: [Claude Fable 5] Good catch — fixed in dd75ea1: the README task list now shows the 3.x names (publish{Alpha/Production}ReleaseApps/...Apk/...Listing, promote...Artifact, verified against ./gradlew tasks --group publishing) and drops the commented-out beta flavor. Thanks.

app/build.gradle
```diff
@@ -142,14 +158,30 @@ android {
...
+    onVariants(selector().all()) { variant ->
+        if (variant.flavorName == "alpha") {
+            variant.outputs.each { output ->
+                output.versionName.set(output.versionCode.map { code ->
+                    "${code.intdiv(100000)}.${(code % 100000).intdiv(1000)}.${code % 1000}-alpha"
+                })
+            }
+        }
+    }
+}
```
andrew-polk, line 175: [Devin] **Investigate**: Alpha versionName sync only verified for the non-bump case

The whole fix hinges on `output.versionName.set(output.versionCode.map { ... })` (app/build.gradle) reconstructing the name from the AUTO-resolved versionCode. Crucially, this only diverges from the old `versionNameSuffix "-alpha"` behavior when AUTO actually bumps the code. In the PR's cited verification (3.4.42-alpha / 304042) AUTO did NOT bump, so the reconstructed name equals the suffix-based name it replaces — meaning that end-to-end check would pass identically even if the reconstruction silently never applied (e.g. if `variant.flavorName == "alpha"` never matched, or the provider chain resolved to the derived code). The genuinely new behavior — a rebuild with no commit where AUTO floors to Play-highest+1 and the name must follow the bumped code — is exactly what the PR notes is not yet proven and requires a real commit-less re-dispatch. Recommend confirming the reconstructed versionName tracks a bumped versionCode before relying on output-metadata.json for the artifact name.

andrew-polk: [Claude Fable 5] Fair challenge on the tautology — so I split the claim in two and tested the half that's testable without Play. The 'silently never applied' branch is now ruled out empirically: temporarily changing the reconstruction to emit a marker ('+probe') and rebuilding produced versionName '3.4.43-alpha+probe' in output-metadata.json — a string the old versionNameSuffix path cannot generate — proving the flavorName match hits, the map closure runs, and its output (derived purely from output.versionCode) is what AGP bakes into the build. The remaining branch — GPP AUTO feeding a *bumped* code into that same provider — is plugin-documented behavior that can only be observed against the real Play API, and it is exactly the PR's planned post-merge verification (dispatch with no new commit, confirm highest+1 with matching name). Worst case if that ever failed is a mislabeled artifact name, not a wrong publish. Resolving; the post-merge dispatch settles the last branch.

andrew-polk:
[Claude Fable 5] Consulted Devin on 2026-07-16 up to commit dd75ea1 (five rounds total across 13505ef → dd75ea1).

Round 4 (the gradle-play-publisher 3.13 upgrade) raised two flags: [stale 2.x task names in the README](https://github.com/BloomBooks/BloomReader/pull/330#discussion_r3599060472) (real — fixed in dd75ea1) and [whether output-metadata.json reflects the AUTO-resolved code](https://github.com/BloomBooks/BloomReader/pull/330#discussion_r3599059945) (mechanism verified; resolved). Round 5 raised [one further flag on the versionName sync](https://github.com/BloomBooks/BloomReader/pull/330#discussion_r3599110268), whose 'silently inert wiring' branch was ruled out with a local marker probe; the remaining branch is the PR's documented post-merge verification. No bugs in any round; all flag threads are documented and resolved.

